PR #2, opened 2019-03-20
remove main.rs and Cargo.lock as sophia is a library

MarcAntoine-Arnaud:


pchampin:
Thanks for the PR.
Unrelated, but here's a "pro tip": **never** use your branch `master` for a pull request, always create a dedicated branch. If I had refused the PR, your `master` and my `master` would have diverged, which would make it hard for you to pull further commits...

Issue #8, opened 2019-05-21
Move from `error-chain` to `err-derive` of `snafu` for error management


althonos:
Hi!

`error-chain` [is not maintained anymore](https://github.com/rust-lang-nursery/failure/issues/181), so it would be a good idea to move to [`failure`](https://github.com/rust-lang-nursery/failure), which has all the features needed already, as well as backtrace support (i.e. the underlying `pest` error could be accessed from a `ParserError` using the `cause()` method).

pchampin:
Thank you for this suggestion. I need more input before I make a decision.

You argue that `error-chain` [is not maintained anymore](rust-lang-nursery/failure#181), however the conclusion of the thread you cite is [not as clear-cut](https://github.com/rust-lang-nursery/failure/issues/181#issuecomment-421734025). Furthermore, it seems that `error-chain` [is still being maintained](https://github.com/rust-lang-nursery/error-chain/commits/master) after all.

That being said, after a quick glance at `failure`, I must say it does look more elegant and lightweight than `error-chain` (I'm not a big fan of the `Error`/`ErrorKind` dichotomy... it feels like coding in Java :smiling_imp:).

Still, I'm a bit uncomfortable with investing in a crate whose ambition is to deprecate a trait from the standard lib. Especially when some of the complains about the standard trait have been fixed in the meantime. In that respect, the [Evolution section in their README](https://github.com/rust-lang-nursery/failure#evolution) is not really clear, but it makes me nervous.

So, could you please elaborate more why you think `failure` is a safer bet for the future than `error-chain`?

althonos:
As you said, `failure` has the advantage of removing the `Error`/`ErrorKind` duplication. Furthermore, it can create a proper backtrace, without ambiguity, from any error, and not just `ChainedError` implementors: the backtrace is reached through the `cause` method, which returns the value of one of the variant's fields. It is unclear how to do that with `error-chain`, since I'm not sure you can both expose an error as an attribute *and* access it through the backtrace.

The other advantage is that the conversion traits make it easier to use `?`, whereas `error-chain` sometimes require you to wrap/convert the error yourself (see the PR for the XML parser, where implementing another error category makes it more verbose to return a top-level error from the lower-level one).

pchampin:
> the backtrace is reached through the cause method, which returns the value of one of the variant's fields

[...]

PR #9, opened 2019-05-22
Work-in-progress RDF/XML parser

althonos:
Hi!

This is a **work in progress** branch, so it's not clear to merge right now, but this is how far I am currently with an RDF/XML parser. I'm parsing correctly ~~most~~ **all** of the [RDF/XML 1.1](https://www.w3.org/TR/rdf-syntax-grammar/#parseTypeCollectionPropertyElt) examples, and I need to add the tests from `rdf-test`.

I'll also feature-gate the parser behind an XML feature since it requires an additional dependency (`quick-xml`).

## Missing features

- [x] Proper error reporting (currently everything panics, but I'm not sure I want to work on this before #8  is resolved)
- [ ] `rdf:parseType="Literal"` (this is likely to need a PR in `quick-xml`)
- [x] `rdf:parseType="Collection"`
- [x] Statements reification
- [x] Validation of elements names


sophia/src/parser/xml.rs
```diff
@@ -0,0 +1,1154 @@
...
+    }
+
+    fn expand_attribute(&self, attr: &str) -> Result<Term<F::TermData>> {
+        if let Some(separator_idx) = attr.chars().position(|c| c == ':') {
+            let prefix = &attr[..separator_idx];
+            let reference = &attr[separator_idx + 1..];
+            if let Some(ns) = self.ns.get(prefix) {
+                ns.get(self.factory.borrow_mut().get_term_data(reference))
+            } else {
+                panic!("unknown namespace: {}", prefix)
```
MarcAntoine-Arnaud: return `Err()` instead of `panic!` ... (apply many time in this file)

sophia/Cargo.toml
```diff
@@ -19,6 +19,7 @@ language-tag = "0.9.0"
 lazy_static = "1.2.0"
 pest = "2.1.0"
 pest_derive = "2.1.0"
+quick-xml = "0.14.0"
```
MarcAntoine-Arnaud: Do you have already tested the [YaSerde](https://github.com/media-io/yaserde/) library ?
It deserialize/serialize struct/enum into/from xml. As Serde but dedicated to XML with specific parameters to mention attributes, namespaces.
But in this case I think it will be complex to use as the model need to be generic (not limited to a struct/enum description).

althonos: The issue of `YaSerde` (as well as any XML DOM library) here is that it would not allow streaming triples, whereas my implementation does (you don't have to wait for the whole document to start producing triples).

sophia/src/parser/xml.rs
althonos: Yes I know, this is WIP, and ultimately all errors will be added to `XmlParser.triples`. I will need to edit the current errors to allow wrapping XML and decoding errors, but that may even be easier with `failure` (as I suggested in #8).

althonos:
cc @phillord if you want to see where this is going

Tpt:
Just to make sure you are aware: rudf provides a simple RDF/XML parser: https://github.com/Tpt/rudf/blob/master/lib/src/rio/xml.rs

phillord:
@Tpt Also got this now https://github.com/phillord/raptor-rs

althonos:
@Tpt : mine however supports `parseType="collection"` :wink:

althonos:
Finally, this a feature-complete ! I still need to do a bit of refactoring, in particular to reduce code duplication and complexity, but this version behaves correctly against the RDF/XML test suite, including for errors (it fails where a failure is expected). The only exception is the `parseType="Literal"` feature, which is not supported by the underlying `quick-xml` library, but I opened an issue in there to request that.

Streaming (i.e. iterating over the produced triples and unwraping the result) the Gene Ontology ([go.owl](http://geneontology.org/docs/download-ontology/#go_obo_and_owl)) takes about 5~10 seconds on my machine, but there is probably still some optimisations to be carried out.

sophia/src/error.rs
```diff
@@ -3,6 +3,9 @@
 use pest::error::{InputLocation, LineColLocation};
 
 error_chain! {
+    links {
+        XmlError(crate::parser::xml::error::Error, crate::parser::xml::error::ErrorKind);
```
pchampin: My idea was rather that all parsers would "raise" `ParserError`, possibly *chaining* their own specific error type.

sophia/src/ns.rs
```diff
@@ -85,6 +86,7 @@ pub mod rdf {
         // classes
         Alt,
         Bad,
+        Description,
```
pchampin: Description is not a class. It should be in the "core syntax terms" list.

sophia/src/ns.rs
```diff
@@ -186,6 +200,44 @@ pub mod rdfs {
...
 
+/// The standard `xml:` namespace
+pub mod xml {
+    namespace!(
+        "http://www.w3.org/XML/1998/namespace#",
+        lang,
+        space,
+        base,
+        id,
+        // John Bosak
```
pchampin:
```suggestion
        // Jon Bosak
```

sophia/src/ns.rs
```diff
@@ -186,6 +200,44 @@ pub mod rdfs {
...
+        Father
+    );
+}
+
+/// The standard `dc:` namespace.
+pub mod dc {
```
pchampin: The `dc` prefix is ambiguous: it is sometimes used for `http://purl.org/dc/elements/1.1/`, and sometimes for `http://purl.org/dc/terms/`. In order to avoid any ambiguity, I prefer to use `dce` or `dc11` for the former, and `dct` for the later.

sophia/src/parser/xml.rs
```diff
@@ -0,0 +1,2071 @@
...
+    rdf::aboutEach,
+    rdf::aboutEachPrefix,
+];
+
+static RESERVED_ATTRIBUTES_NAMES: &'static [StaticTerm] =
+    &[rdf::li, rdf::aboutEach, rdf::aboutEachPrefix, rdf::bagID];
+
+mod xmlname {
+
+    use pest::Parser;
```
pchampin: Couldn't use simply use regex here?

sophia/src/ns.rs
althonos: I was using it for my tests, but ultimately stopped; since it is not used anywhere it could just be removed.

pchampin: works for me :)

sophia/src/parser/xml.rs
althonos: Indeed.

sophia/src/error.rs
althonos: Ah yes, I see what you mean. Unfortunately `quick-xml` is only a streaming parser, so it does not store the line/col location, only the input location...

pchampin: Not a problem: you should use `LineColLocation::Pos(0,0)`.

I know, this should be documented :-/

pchampin:
I know that last commit cost you :wink:, much appreciated.
I'll take over to please Travis, and do the merge. 
Thanks again

althonos:
@pchampin : my only regret is not being able to add better error reports, I should have used `xmlparser`  instead of `quick-xml` to have input spans and better error reporting, I may try to experiment to see how easy it is to replace it ! :wink:

PR #13, opened 2019-08-06
Fix XML parser and bump URL dependency 

### LINKED ISSUE #12: `percent_encoding` becomes private in `url`>=2.0.0
...and the XML parser uses it, which prevents us from upgrading the dependency to `url` beyond 1.7.2. @althanos, could you please have look at it and see if we can get rid of this? Thanks in advance...

althonos:
This PR fixes #12 by adding `percent-encoding` as a direct dependency, and updates the `url` crate to version `2.1.0`. I also added a field to `Cargo.toml` so that the documentation is built with all features enabled on `docs.rs`, so that the XML parser doc is built as well.

By the way, @phillord and I would be glad to have the XML parser released in a new version (`0.3.0` I suppose) whenever you can push that.

### DIFF
--- a/sophia/Cargo.toml
+++ b/sophia/Cargo.toml
@@ -8,8 +8,10 @@ documentation = "https://docs.rs/sophia"
 readme = "../README.md"
 license = "CECILL-C"
 keywords = ["rdf", "linked-data", "semantic-web"]
-
 edition = "2018"
+
+[package.metadata.docs.rs]
+all-features = true
 
 [features]
 default = []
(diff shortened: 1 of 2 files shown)


pchampin:
Thanks, that was fast!
Re. the 0.3.0 release, I'm actually in the process of doing that (at long last), and this is how I came into this version problem. I saw your PR literaly seconds before typing `cargo publish` :-D

phillord:
@pchampin Thanks for the release! Makes it much easier to use.

Issue #14, opened 2019-12-02
Provide some "Getting Started" documentation


JordanShurmer:
This create looks really promising for my Linked Data project. However, I'm really unsure about how to get started using it.

It would be really useful to have some examples of doing the most common things in the rust docs. For example: How to create a graph, add some triples, and write it as turtle to a file; How to read a turtle file in, parse it as a graph, then modify it's contents.

I'd be happy to contribute the documentation, if I can get some pointers on how to do it :)

pchampin:
Thanks for your interest in this crate. And thanks to @MattesWhite for quickly responding with a proposal. I adapted the example and merged it.

I leave this issue open, because I intend to add more examples later. But that will be after I integrate more parsers (Turtle, TriG...) from the https://github.com/Tpt/rio crate.

JordanShurmer:
Awesome! Thanks @MattesWhite

huhn511:
I think it would be nice to have some examples for each crate. Also very helpful, would be a [mdbook](https://github.com/rust-lang/mdBook) for detailed documentation - and the rendered docs page a nice start for developers. 

What do you think about it? I could add PR with a docs folder and prepare something (setup mdbook, structure. GitHub deploy action, etc).

Best wishes ✌️

pchampin:
@huhn511 That would be really nice indeed, and yes, a PR would be greatly appreciated :)

hoijui:
In a loop over all triples, I wanted to put all literal objects in one vec, all subjects and non-literal (and non-blank) objects in one vec, and all predicates + their subject and object in an other data structure (vec?), to be able to use them in a graph-data structure ([petgraph](https://docs.rs/petgraph/0.5.1/petgraph/)) later on. I was not able to do that, as I am a rust newb, and got lost in the lifetime, and nested data-structure complexity of it (from the sophia side).
Such an example would be nice to see for me (without the petgraph part, maybe?).

hoijui:
I guess I woudl store the `Rc<str>` in my `vec`, right?
how would I get to it, in the loop?

pchampin:
@hoijui I just opened the [discussion](https://github.com/pchampin/sophia_rs/discussions) feature, so that we can have StackOverflow style Q&A's, which I think is more appropriate for that kind of question. I suggest that you open a question there, and provide an code example... I'll have a look.

hoijui:
done: https://github.com/pchampin/sophia_rs/discussions/102

[...]

PR #15, opened 2019-12-02
Add a getting started example

MattesWhite:
Added a 'Getting Started' example to the crate's docs as requested by #14.

### DIFF
--- a/sophia/src/lib.rs
+++ b/sophia/src/lib.rs
@@ -25,9 +25,36 @@
 //! they will be called, respectively,
 //! the *generalized* RDF model, and the *strict* RDF model.
 //!
-//! # Examples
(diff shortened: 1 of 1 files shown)


MattesWhite:
@pchampin 
The build fails because `Trait objects` with no `dyn` keyword where recently depricated and now raise warnings. Should I correct these with this PR or will this be solved soon?

pchampin:
@MattesWhite thanks for this PR. I made some improvements to make it more "idiomatic" in 9079052. 
Also, I had a very strange GIT error, which is why I didn't reuse your commit above.

pchampin:
@MattesWhite (sorry I didn't notice your previous message)
Yes, I intend to clean this soon.

Issue #24, opened 2019-12-10
Get rid of the lifetime parameters on Triple, Graph and Dataset


pchampin:
They are causing much complexity, especially the [requirement for using higher-ranked type bound](https://docs.rs/sophia/0.3.0/sophia/graph/trait.Graph.html#how-to-use-graph-in-a-trait-bound) in some situations.

It might be better to give up some genericity in order to make the API easier to use.

pchampin:
Removing lifetimes on Tripe and Quad is easy. I have a local branch that I'll push soon, where this is done.

For Graph and Dataset, this is more tricky. The problem comes from the fact that Graph::Triple, which is the type returned by the iterating methods (triples, triples_with_s, etc.) sometimes needs a lifetime, especially when one wants the iterator to return *references* to internal triples owned by the graph. This is something that I would like to keep.

As mentioned [elsewhere](https://github.com/pchampin/sophia_rs/issues/23#issuecomment-563998845_), the cleanest solution would be generic associated types (GAT), but those are not expected to be stabilized before long...

I have experimented with another approach, which tries to emulate somehow GAT, by using pointers under the hood, hence eschewing lifetime parameters, and wrapping them in a safe abstraction. I tried my best to make this approach easy enough for implementors of the Graph trait (although the documentations is still rough at the edges).

https://gist.github.com/pchampin/7fbf3262ab22b6a0f6a6eae8a6aa938c

@Tpt @MattesWhite anyonelse... your feedback would be much appreciated

MattesWhite:
Your solution seems suitable.

* One minor disadvantage I see is the cost of extra wrapping of newly created triple but maybe the compiler is smart enough to optimize that away.
* What definitley must be document is that this is just a temporary solution until GATs arrive.
* A small suggestion I have to make: When implementors see that they have to return `GuradedTriples` they will probably have first a look at the docs of the type. As a consequence, I would move the constructors to the type itself.
* Your implementation exposes the raw pointers to the user/implementor. [Here](https://gist.github.com/MattesWhite/f5daf2d38d36c4e325e00cbb41ec59e8) is a version with a complete safe API and without `UnsafeTriple`.

Tpt:
I like @MattesWhite solution. A maybe dumb comment: are the NotNull pointers actually required in your solution? I believe you could do it with just references (but maybe I'm wrong).

MattesWhite:
[...]

PR #29, opened 2019-12-16
Replace coercible errors

MattesWhite:
This PR replaces the `coercible-errors`-crate by `StreamError`, defined like following:
```Rust
#[derive(Debug)]
pub enum StreamError<SourceErr, SinkErr>
where
    SourceErr: 'static + Error,
    SinkErr: 'static + Error,
{
    Source(SourceErr),
    Sink(SinkErr),
}
```
It allows function to return two different types of error, usually either an error from the `Source` providing new data or from the `Sink` that consumes it.

### to do

As showed are the trait bounds for `SourceErr` and `SinkErr` only `'static + Error`. This should be expanded by `+ Send + Sync` in the future, in order to be able to implement concurrent graphs. The bound extension is delayed until `sophia::error::Error` is completly removed.

_This PR is the result of #8 and was implemented according to #28_


MattesWhite:
Okay I have no idea why Travis failed. The log seems like it has built an old version of the parser-module? Did I something wrong at commiting?

pchampin:
I'm assuming you didn't use the `--all-features` option when you ran the tests locally. So you are missing the feature-gated parts of the code (which, it seems, still refer to `coercible-error`).

I'm about to make a reviw of your PR, so you might want to wait for my comments before putting further work in iit. All in all, this is good, but there are a few changes that I'd like to suggest.

sophia/src/triple/stream/_error.rs
```diff
@@ -0,0 +1,214 @@
...
+/// The `std::error::Error::source()` method is implemented accordingly.
+///
+/// # ToDo
+///
+/// For a proper use in multi-threaded environments the trait bounds should be
+/// supplied with `+ Send + Sync`. However, such bounds wound make `StreamError`
+/// incompatible with `sophia`'s whole error-handling. This should be resolved
+/// when the error-handling is completly refactored
+/// ([tracking issue](https://github.com/pchampin/sophia_rs/issues/8)).
+#[derive(Debug)]
```
pchampin: Why not use `thiserror` already? This is the plan eventually, and this would reduce the number of lines of code...

sophia/src/triple/stream/_error.rs
```diff
@@ -0,0 +1,214 @@
...
+pub enum StreamError<SourceErr, SinkErr>
+where
+    SourceErr: 'static + Error,
+    SinkErr: 'static + Error,
+{
+    Source(SourceErr),
+    Sink(SinkErr),
```
pchampin: I think we should rename the variants `SourceError` and `SinkError`. This is a bit redundant with the name of the enum, but that makes them suitable for public export (so anyone, even outside this crate, can use `SourceError` rather than the more verbose `StreamError::Source`).

sophia/src/triple/stream/_error.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+use self::StreamError::*;
```
pchampin: This would then become `pub use ...`

sophia/src/triple/stream/_error.rs
```diff
@@ -0,0 +1,214 @@
...
+    SinkErr: 'static + Error,
+{
+	/// Build a `StreamError` from an error of the `Source`.
+    pub fn from_source_err(se: SourceErr) -> Self {
+        Source(se)
+	}
+	/// Build a `StreamError` from an error of the `Sink`.
+    pub fn from_sink_err(se: SinkErr) -> Self {
+        Sink(se)
+    }
```
pchampin: I'm really not sure we need this if we publicly expose SourceError and SinkError...
`SourceError(err)` is much easier than `StreamError::from_source_err(err)`

sophia/src/triple/stream/_error.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// Checks if `StreamError` was raised by the `Source`.
+    pub fn is_source_err(&self) -> bool {
```
pchampin: for the sake of homogeneity, this should probably become `is_source_error`
(same for `is_sink_err` below)

sophia/src/triple/stream/_error.rs
```diff
@@ -0,0 +1,214 @@
...
+	/// Converts the ``StreamError` into its `SourceError`.
+	/// 
+	/// _Note_ this also works if `SinkError` is `Infallible`.
+    pub fn into_source_err(self) -> SourceErr {
+        match self {
+            Source(err) => err,
+            Sink(err) => err.into(),
+        }
+    }
+}
```
pchampin: I think this impl and all the following similar ones could be replaced by a *single* method, in the main impl block, that would look like this:
```rust
    pub fn unwrap_into<Err>(self) -> Err where
        Err: From<SourceErr> + From<SinkErr>
    {
        match self {
            SourceError(err) => err.into(),
            SinkError(err) => err.into(),
        }
    }
````
This would make the API of `StreamError` more simple and homogeneous.

Furthermore, the naming of the methods is misleading. At first I thought that `into_source_err` was meant to change the Sink variant into the Source variant, which I thought was odd...

File: sophia/src/triple/stream/_error.rs [hunk 7]
```diff
@@ -0,0 +1,214 @@
...
+///
+/// The counter part is [`SinkResultExt`](trait.SinkResultExt.html)
+pub trait SourceResultExt<T, SourceErr>
+where
+    SourceErr: 'static + Error,
+{
+    fn map_source_err<SinkErr>(self) -> Result<T, StreamError<SourceErr, SinkErr>>
+    where
+        SinkErr: 'static + Error;
+}
```
pchampin: I'm really not sure we need this (nor the following) extension trait.
`my_result.map_source_err()` could be written `my_result.map_err(SourceError)`,
which is much more idiomatic and hardly more verbose.

sophia/src/dataset/_traits.rs
```diff
@@ -714,10 +699,11 @@ pub trait MutableDataset: for<'x> Dataset<'x> {
                     q.g().map(BoxTerm::from),
                 )
             })
-            .collect::<std::result::Result<_, _>>()
+            .collect::<std::result::Result<Vec<_>, _>>()
             .map_err(Into::into)?;
         let mut to_remove = to_remove.into_iter().as_quad_source();
-        self.remove_all(&mut to_remove)?;
+        self.remove_all(&mut to_remove)
+            .map_err(|err| err.into_sink_err())?;
```
pchampin: according to my suggestion on `StreamError`, this would become
```suggestion
            .map_err(|err| err.unwrap_into())?;
```
I'm guessing the correct type will be inferred.

sophia/src/graph/_traits.rs
```diff
@@ -481,35 +472,33 @@ pub trait MutableGraph: for<'x> Graph<'x> {
...
                 ]
             })
-            .collect::<std::result::Result<_, _>>()
+            .collect::<std::result::Result<Vec<_>, _>>()
             .map_err(Into::into)?;
         let mut to_remove = to_remove.into_iter().as_triple_source();
-        Ok(self.remove_all(&mut to_remove)?)
+        Ok(self
+            .remove_all(&mut to_remove)
+            .map_err(|err| err.into_sink_err())?)
```
pchampin: Same as above
```suggestion
            .map_err(|err| err.unwrap_into())?)
```

sophia/src/graph/_traits.rs
```diff
@@ -519,10 +508,11 @@ pub trait MutableGraph: for<'x> Graph<'x> {
                     BoxTerm::from(t.o()),
                 ]
             })
-            .collect::<std::result::Result<_, _>>()
+            .collect::<std::result::Result<Vec<_>, _>>()
             .map_err(Into::into)?;
         let mut to_remove = to_remove.into_iter().as_triple_source();
-        self.remove_all(&mut to_remove)?;
+        self.remove_all(&mut to_remove)
+            .map_err(|err| err.into_sink_err())?;
```
pchampin: same as above
```suggestion
            .map_err(|err| err.unwrap_into())?;
```

sophia/src/parser/nq.rs
```diff
@@ -263,7 +264,7 @@ mod test {
         let mut d = HashSetDataset::new();
         let reader = io::Cursor::new(GENERALIZED_DOC);
         let res = STRICT.parse_read(reader).in_dataset(&mut d);
-        if let Err(Error(ParserError(_, location), _)) = res {
+        if let Err(StreamError::Source(Error(ParserError(_, location), _))) = res {
```
pchampin: Could be simplied as 
```suggestion
        if let Err(SourceError(Error(ParserError(_, location), _))) = res {
```
if you accept my suggestions on `StreamError`

sophia/src/quad/stream.rs
```diff
@@ -63,22 +60,19 @@ impl<'a, I, T, E> QuadSource for I
...
         sink: &mut TS,
-    ) -> CoercedResult<TS::Outcome, Self::Error, TS::Error>
-    where
-        Self::Error: CoercibleWith<TS::Error>,
-    {
+    ) -> Result<TS::Outcome, StreamError<Self::Error, TS::Error>> {
         for tr in self {
-            let t = tr?;
-            sink.feed(&t)?;
+            let t = tr.map_source_err()?;
```
pchampin: As suggested in `StreamError`, this is uselessly unidiomatic.
I would rather write
```suggestion
            let t = tr.map_err(SourceError)?;
```
and do the same with `SinkError` below

sophia/src/triple/stream/_error.rs
MattesWhite: That's a pretty good idea. I'll go with that.

sophia/src/triple/stream/_error.rs
MattesWhite: I prefer to call the function `into_inner()`. For me 'unwrap' always signals some operation that might _panic_ if not called properly. Which this function clearly not does.

pchampin: Granted, `unwrap` was not the best choice.
Shall I be over-pendantic and suggest `inner_into`, because we are not exactly converting the StreamError into its inner error, we are rather converting the inner error into potentially something else...

pchampin:
@MattesWhite is this ready for merge? (guessing so, but just checking)

MattesWhite:
Yes its ready 👍

pchampin:
Ok, I'll try to do it during the week-end. And I'll also try to push and merge my work on #24. The latter is impacting almost all files in the repo, so I advice you don't start any work until it's done -- merging it afterwards would be hell.
Thanks again for your help.

Issue #25, opened 2019-12-10
Provide a simpler parser API


MattesWhite:
This issue contributes to #23 

For me the parser API of `sophia` is a bit confusing with its macros and `Config`.

I suggest the introduction of a `Parser`-trait:
```Rust
pub trait Parser {
    fn parse_str<'src>(&'src mut self, input: &'src str) -> 
        Result<Box<dyn Iterator<Item=Result<[Term<Cow<'src, str>>; 3]>> + 'src>>;
    fn parse_read<'src, R: 'src + Read>(&'src mut self, input: &mut R) -> 
        Result<Box<dyn Iterator<Item=Result<[Term<Cow<'src, str>>; 3]>> + 'src>>;
}
``` 
_For `Quad`s a similar trait respectively_

* Allows `Parser`s to store state so they can lazily parse.
* General enough for all serialization formats in question.
* Easy for beginners to understand.
* Result implements `TripleSource`

An implementation for the N-Triples parser, e.g.:
```Rust
mod nt {
  #derive[default]
  struct Parser {
    strict: bool;
  }

  impl Parser {
    fn new_strict() -> Self { Self { strict: true, } }
  }
 
  impl Parser for Parser {
    ...
  }
}
```

pchampin:
Re. traits vs. macros, big +1. Initially, I wanted to define a `Parser` trait, but was unabled to complete it and opted instead for the "abstract interface" + macros approach, but I am not entierly satisfied with it. So thanks for prompting me, I'll give it a try again.

Leaving aside the trait vs. macro problem, what you propose above is actually very close to the current API, expect that what you call `Parser` is currently called `Config`. And this was a deliberate choice, because the actual *parser*, in my view, is only created when you call `parse_str` or `parse_read`. You mention "storing state so they can lazily parse" (which I think is what current implementations are doing), but such state (e.g. prefix bindings in Turtle) should not be shared across multiple calls of `parse_X`. The only thing that is really persistent here is, therefore, the parsing *configuration*, hence the current name.

That being said, I agree that this subtle semantic distinctions may bring more confusion than clarity. So we might as well bite the bullet and use the name everyone expects.

To sum it up: my plan is to replace the current `Config` "protocol" by a `Parser` trait, offering roughly the same interface.

pchampin:
The discussions about #8 may have impact on this issue. We actually have two options:

[...]

PR #31, opened 2020-01-09
Refactor error handling in mods graph and dataset

MattesWhite:
Continuation of #8 .

Removes unnecessary error variants from `sophia::error::Error` and adds a new error for the `Graph` to `Dataset` wrapper.

### Future work

At investigating `GraphAsDataset<G, H>` I recognized that the second parameter `H` is used nowhere actually. I could remove this within this PR if you like or did I missed something?

@pchampin It seems to me that the mods parser and serializer are currently adapted to the `rio` crates. Is this work finished and I can continue refactoring the error handling or should I wait?

### DIFF
--- a/sophia/src/error.rs
+++ b/sophia/src/error.rs
@@ -5,14 +5,6 @@ use std::fmt;
 
 error_chain! {
     errors {
-        /// Raised by the methods of the [`Graph`](../graph/trait.Graph.html) trait.
-        GraphError(message: String) {
-            display("error while querying Graph: {}", message)
-        }
-        /// Raised by the methods of the [`MutableGraph`](../graph/trait.MutableGraph.html) trait.
-        GraphMutationError(msg: String) {
-            display("error while modifying Graph: {}", msg)
-        }
         /// Raised whenever an invalid prefix is used in a PName.
         InvalidPrefix(prefix: String) {
             display("invalid prefix <{}>", prefix)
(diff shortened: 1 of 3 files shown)


pchampin:
> At investigating GraphAsDataset<G, H> I recognized that the second parameter H is used nowhere

Actually, it *is* used. GraphAsDataset does not wrap a G(raph) but a Borrow<G> (which is the trait bound on H). So you can decide to wrap either an owned graph or a simple reference to a graph.

> It seems to me that the mods parser and serializer are currently adapted to the rio crates.

Not sure what makes you think that... I tried to keep rio-specific code in rio_common.rs and all the parsers that are now based on rio (nt, nq, turtle, trig, ntrig), but parser.rs and serializer.rs are meant to be agnostic. If they are not, this is a bug.

> Is this work finished and I can continue refactoring the error handling or should I wait?

I won't be able to work on it in the next week or two, so you are free to go. That being said, may be it would make sense to wait until theses modules are refactored to expose proper traits (as you suggested in #25)...

pchampin:
Apart from all that, I like what's in this PR. Do you want me to merge it now, or do you plan to include the symetrical work in `dataset` before? (I'm asking because the title of this PR is hinting at it, not to force you into more work... :sweat_smile:)

MattesWhite:
Actually the work for `dataset` *is* done. Every existing implemtation of `Dataset` uses `Infallible` as error (except `GraphAsDataset` 😉) so far. Therefore, there is nothing to be done.

pchampin:
Ok cool, merging at once.

After all, I *am* working now on the parser module, so you might want to wait a little bit before touching it...

PR #39, opened 2020-02-14
Correct paths in namespace macro

MattesWhite:
When the `namespace!` macro is used like this (2018 edition):

```rust
use sophia::namespace;

namespace!(...);
```

instead of using

```rust
#[use_macro]
extern crate sophia;

namespace!(...);
```

The macro throws an error because the `ns_term!` is not exported.

This PR fixes this issue by adding the absolute path to macro-calls inside the macros.


pchampin:
Thanks :)

PR #41, opened 2020-02-24
New namespace "owl:"

tbourg:
# New `owl:` namespace
-------------------------------------
The rasoner needs the `owl:` namespace


pchampin:
Nitpicking a bit, but would you mind sorting the terms in the namespace?
Alphabetical order would be the most natural, perhaps putting all "classes" first, then all "properties"...

PR #43, opened 2020-02-25
Minor changes to serializers

MattesWhite:
Response to #34 commit  9a97dae .

- Remove `&mut` from source  in serialize 'statement' methods.
  Unnesessary as sources are completely consumed (see doc: 'Serialize __all__ triples/quads').
- Remove `NxStringifier` type alias.
   Seems redundant to call `NtStringifier::new_stringifier()` when
   `NtSerializer::new_stringifier()` does the same.
- Rename `NtSerializer` to `NqSerializer` in nq.rs.

### DIFF
--- a/sophia/src/serializer.rs
+++ b/sophia/src/serializer.rs
@@ -28,7 +28,7 @@ pub trait TripleSerializer {
     /// [`TripleSource`]: ../triple/stream/trait.TripleSource.html
     fn serialize_triples<TS>(
         &mut self,
-        source: &mut TS,
+        source: TS,
     ) -> StreamResult<&mut Self, TS::Error, Self::Error>
     where
         TS: TripleSource,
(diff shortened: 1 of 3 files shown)


pchampin:
Thanks for this proposal. Comments below

* Consuming Triple/Quad Source → I'm not entirely sure... see below.
* Removing NxSerializer alias → good call, this is indeed better that way
* Renaming NtX to NqX → of course, silly me :-/


Most (if not all) places where Triple/Quad Sources are used (e.g. `Graph::insert_all`), they are borrowed mutably. The reason for not consuming them is that some sources (especially parsers) may still carry valuable information after they have exhausted their triples (e.g. the prefix declaration they have encountered). See #45 for a discussion on that topic.

Would you mind making a new PR with only the last two items, and wait till we resolve #45 for the first one? Thanks

MattesWhite:
Reduced version committed in #46

PR #47, opened 2020-02-26
Variable as own type

MattesWhite:
Contribution according to the discussions in #34 and #35.

As a first step this moves variables out of `Term` and into an own type `Variable`.

Following features are implemented:

- (unsafe) constructor
- copy with factory
- write to fmt/io
- convert between `Term` and `Variable` back and forth
- Integration of `Variable` in the methods of `Term`
- own public submodule `sophia::term::variable`

---

Step by step each kind of RDF term will get its own public submodule. I started with variables due to their small codebase. 
I think it is better to continue this process in small PRs rather then working weeks on an own branch and resolving ocuring conflicts in the end.

### DIFF
--- a/sophia/Cargo.toml
+++ b/sophia/Cargo.toml
@@ -30,3 +30,6 @@ thiserror = "1.0.9"
 
 quick-xml = { version = "0.17.2", optional = true }
 percent-encoding = { version = "2.1.0", optional = true }
+
+[dev-dependencies]
+test-case = "1.0.0"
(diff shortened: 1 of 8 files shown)


pchampin:
Looks like cargo fmt is till unhappy...

pchampin:
> Somehow I managed to butcher the latest merged PR #46.
> Repaired with this commit.

Yuck! A `rebase` and a `push --force` would have been cleaner.

Instead, I cherry-picked the relevant commits from this PR (squashing the two fmt-related ones), and merged them in 19e854b1. You are still credited as the authors of the new commits.

Thanks for this work, I like where this is going.

PR #48, opened 2020-02-27
Make BlankNode a public, own type

MattesWhite:
Continuation of turning kinds of term into own, public types (see #34 )

- turned mod `_bnode_id` into pub mod `blank_node`
- Refactored `BNodeId` to new-type struct `BlankNode`
- Aligned `BlankNode` interface with newly created `Variable`
- `BlankNode::new()` and `Term::new_bnode()` now fail correctly if not N3 _[1]_
  - Adapted tests accordingly
- Integrated `BlankNode` into `Term`
- Minor corrections in mod `variable`

_[1]_ This is a significant change but this is more aligned to the way variables are handled.

---

I think the next parts IRI and literal will take longer as they are way more complex as variables and blank nodes (in terms of implementation). Probably, I will create a WIP-PR so you can watch the progress and we can discuss certain changes.


sophia/src/term/_display.rs
```diff
@@ -29,12 +29,16 @@ where
...
+            if bn.is_n3() {
+                bn.write_fmt(w)?;
             } else {
-                write_non_n3_bnode_id(w, ident.as_ref())?;
+                // non conformant identifier
+                w.write_str("_:_")?;
+                for b in bn.as_ref().as_bytes() {
+                    write!(w, "{:x}", b)?;
+                }
+                w.write_str("_:_")?;
```
MattesWhite: Is this still necessary? With this PR it is already checked at construction if a blank node is valid according to N3, therefore, there should no longer be non-conformant blank nodes.

pchampin: Agreed. At some point I wanted to be more lenient with bnode identifiers, but that adds complexity, and is inconsistent with how we handle variables.

sophia/src/term/blank_node.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    /// Whether this blank node identifier is compatible with the N3 family
+    /// of syntaxes (N-Triples, Turtle...).
+    ///
+    /// This is already checked by the standard constructor
+    /// [`new()`](struct.BlankNode.html#mehod.new).
+    pub fn is_n3(&self) -> bool {
+        BLANK_NODE_LABEL.is_match(self.as_ref())
+    }
```
pchampin: This method is moot now that we block non-N3 identifier in the constructor. Probably good to remove it completely.

sophia/src/term/test.rs
```diff
@@ -438,13 +426,14 @@ fn term_similar_but_not_eq() {
     let txt = "http://champin.net/#pa";
     let t1 = StaticTerm::new_iri(txt).unwrap();
     let t2 = StaticTerm::new_literal_dt(txt, xsd::anyURI).unwrap();
-    let t3 = StaticTerm::new_bnode(txt).unwrap();
+    // "http://champin.net/#pa" is not a valid blank node identifier
+    // let t3 = StaticTerm::new_bnode(txt).unwrap();
```
pchampin: Correct. I suggest changing txt to "foo" (which is a valid relative IRI ref), so that we can actually have similar terms of all categories.

Would you mind also adding t4 as a variable, for the sake of completeness?

sophia/src/term/variable.rs
```diff
@@ -5,18 +5,22 @@
...
+use std::convert::TryFrom;
 use std::fmt;
 use std::io;
 
 lazy_static! {
     /// Production of SPARQL's VARNAME according to the
     /// [SPARQL spec](https://www.w3.org/TR/sparql11-query/#rVARNAME).
     ///
-    /// _Note:_ This regular expression matches the whole input (`^...$`),
+    /// # Cpatures
```
pchampin:
```suggestion
    /// # Captures
```

sophia/src/term/blank_node.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+impl<TD> Hash for BlankNode<TD>
+where
+    TD: TermData,
+{
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.as_ref().hash(state)
+    }
+}
```
pchampin: This can actually be replaced by a simple `derive Hash`. Should have been so from the start.

sophia/src/term/blank_node.rs
MattesWhite: You're right I removed it.

PR #49, opened 2020-03-01
IRI as own public type

MattesWhite:
This PR refactors `IriData` into the first-class kind of RDF term `Iri`.

As this affects many areas of `sophia` this PR is marked as work-in-progress and is not ready to be merged, yet.

The changes will not only include 1-to-1 translations of existing code but also re-organization of modules, adding and removing of features and alignment with the previous work of refactoring blank nodes (#48) and variables (#47). As I'm not the owner of this crate and due to the reach of this PR, I highly appreciate reviews and discussions.

**ToDo list**

This list is of course open to discussion. I'll try to work through the list from top to bottom.

- [x] refactor `_iri_data` into a public module, including renaming `IriData` to `Iri`.
- [x] Integrate the `join` mechanism more deeply into `Iri`, as far as possible.
  - [x] write an adapter for iterators, e.g. `graph.triples().resolve_with(&term)...`.
- [x] Integrate new `Iri` into `Term`, including convenient functions.
- [x] ~Replace usage of `Term` where actually only an `Iri` is valid.~ Was not necessary.
- [x] ~Replace `Option<&Term>` with~ Seems not to give any advantages.
  ```rust
  enum GraphId<'a, TD> {
      Named(&'a Iri<TD>),
      Default,
  }
  ```


pchampin:
>  Integrate the join mechanism more deeply into Iri, as far as possible.

+1

> write an adapter for iterators, e.g. graph.triples().resolve_with(&term)....

I can see your use case here, but I think I would rather go for something like
```
    graph.triples().map_triples(|t| base.resolve_triple(t))
```
(see #50 for my thought on the issue of resolving relative IRIs).

> Integrate new Iri into Term, including convenient functions.

+1

> Replace usage of Term where actually only an Iri is valid.

What do you have in mind. Recall that Sophia aims to support a [generalized model](https://docs.rs/sophia/0.3.0/sophia/#generalized-vs-strict-rdf-model). In particular, any term is allowed in any position of Triples/Quads.

> Replace Option<&Term> for graph names with [a dedicated enum]

Your proposed enum is incorrect w.r.t. the RDF Abstract Model, which allows IRIs **or blank nodes** as graph names. Furthermore, as mentioned above, I'd like to keep the Sophia generalized model as uniform as possible. Literal as graph names are very debatable, but variables as graph names will be useful when we want to represent queries as generalized datasets.

An orthogonal concern is: should we have a dedicated type for graph names, instead of simply using `Option<Term>`? I previously used a dedicated type, and I found it caused much complexity without having real added value. Furthermore, I think the semantics of the graph name in a quad is accurately conveyed by `Option<Term>`: triples from the default graph have *no* associated graph name, while triples from named graphs have *some* associated graph name, which is a term.

So I'm not really favorable to that last item. But I'm still open for discussion...

.gitignore
```diff
@@ -2,4 +2,4 @@
 */target
 **/*.rs.bk
 Cargo.lock
-
+/.vscode
```
pchampin: In general, I'm not for including environment specific rules in the published `.gitignore`. Otherwise, it would rapidly be polluted with all the config directories of all possible editors used by different contributors. I advice that you put instead a `.gitignore` file *inside* your `.vscode` directory, containing `*`.

MattesWhite: Sry this was more an accident as on purpose I'll remove it with the next commit.

MattesWhite:
Regarding the replacement of `Term` where only `Iri` is valid.

Here I had something in mind like `Namespace` or the data type of literals as they really only work with IRIs. At the moment there is error handling or a panic in those cases. I thought it would be nice to let the type system take care of such situations. Of course, such replacements must be evaluated with user-friendliness in mind.

MattesWhite:
In the latest commit I removed `unsafe` from `Iri`'s `*_unchecked()` methods due to the [question I asked](https://users.rust-lang.org/t/unchecked-always-unsafe/38975) at URLO.

sophia/src/term/iri.rs
```diff
@@ -36,11 +36,19 @@ pub enum Normalization {
 ///  - provides some identical methods to what `&str` provides (see below);
 ///  - can otherwise be converted to a `String` with `to_string`;
 ///
-/// See [module documentation](index.html)
-/// for more detail.
+/// # Contract
+///
+/// Each `Iri` represents a valid IRI according to the
```
pchampin:
```suggestion
/// Each `Iri` represents a valid IRI references according to the
```

According to the RFC, an IRI is always absolute; an *IRI reference*
can be either absolute or relative.

pchampin: Note that I'm not suggesting that we change *all* mentions of IRI into "IRI references". That would be overly verbose,  even though strictly-speaking, most of the time we write "IRI" it actually means "IRI reference"...

But at least here, were we quote the RFC, I think we should use the proper term.

sophia/src/term/iri.rs
```diff
@@ -160,7 +160,7 @@ where
         }
     }
 
-    /// The length of this IRI.
+    /// The length in this IRI.
```
pchampin: Conceptually, the IRI is a single string, and this is the len *of* that string...

sophia/src/term/iri/_join.rs
```diff
@@ -131,11 +131,11 @@ impl<'a> IriParsed<'a> {
     }
 }
 
-impl<'a> Resolve<String> for IriParsed<'a> {
+impl<'a> Resolve<&str, String> for IriParsed<'a> {
     /// Resolve an IRI given as `String`.
     ///
     /// Fails if `other` is not a valid IRI.
-    fn resolve(&self, other: &String) -> Result<String> {
+    fn resolve(&self, other: &&str) -> Result<String> {
```
pchampin: `&&str` is ackward... Why not just
```rust
impl<'a> Resolve<str, String> for IriParsed<'a> {
    fn resolve(&self, other: &str) -> Result<String> {
       // ...
    }
}
```
?

MattesWhite: The trait is defined as follows:

```rust
pub trait Resolve<S, T = S> {
    fn resolve(&self, other: &S) -> Result<T>;
}
```

which means that `T` must be `Seized` but `str` isn't so your proposal doesn't work for this definition. But your right it would be nicer with only `str`. I'll give it a try but it'll need some trickery with trait bounds.

sophia/src/term/iri.rs
```diff
@@ -0,0 +1,854 @@
...
+                    Cow::Owned(iri)
+                } else {
+                    // case: neither contains a separator
+                    let new_ns = self.value();
+                    let iri = Iri {
+                        ns: new_ns.into(),
+                        suffix: None,
+                        absolute: self.absolute,
+                    };
+                    Cow::Owned(iri)
```
MattesWhite: There is a test case `"tag:", "foo"` -> `"tag:foo"` in `sophia::term::test::iri_normalized_last_hash_or_slash.

Is this really the intended behavior? Or should in this case the suffix rather stay untouched?

MattesWhite:
I noticed there are a bunch of requested changes to the documentation. Don't think I won't address them. I just wait for a final commit to this PR when it's done.

MattesWhite:
> - [ ] write an adapter for iterators, e.g. `graph.triples().resolve_with(&term)...`.

As you opened a dedicated issue (#50) should a solution be included here or would you rather like a separate PR?

pchampin:
Re #50 this PR is fine: your `Resolve` trait is exactly what I had in mind.

sophia/src/term/iri.rs
pchampin: That's a fair question. Actually, hash_or_slash should probably be replaced by gen-delim (see

MattesWhite: So should it include all `gen-delims`?

```
gen-delims     = ":" / "/" / "?" / "#" / "[" / "]" / "@"
```

or only `/`, `#` and `:`?

MattesWhite:
In the last commit I added  a `resolve_triples()` method to `TripleSource`. If you think the design is okay I would add implementasion for `QuadSource` and an extension trait for `Iterator`s (`Item = Term, Triple, Quad, Result<Term>, ...`).

sophia/src/term/iri.rs
pchampin: I think it would make sense to include all gen-delims.

sophia/src/term/iri.rs
```diff
@@ -331,6 +335,28 @@ where
...
 
+    /// Parses the components of the IRI.
+    ///
+    /// This is necessary to use the IRI as a base to resolve other IRIs.
+    ///
+    /// # Errors
+    ///
+    /// As a regular expression is applied to the internal representation of
+    /// the IRI, it is required that the IRI is not suffixed. Otherwise
+    /// `TermError::IriParse` is raised.
```
pchampin: If prefix ends with a gen-delims, it should in theory be possible to build a consistent IriParsed pointing to parts of the prefix and the suffix. This is not technically trivial though, and should not be our priority here. But maybe worth submitting an issue (tagged with 'improvement')?

sophia/src/term/iri.rs
```diff
@@ -375,7 +417,7 @@ impl Iri<&'static str> {
     /// # Safety
```
pchampin: Should this section be renamed, since the function is not `unsafe` anymore?

sophia/src/term/iri.rs
```diff
@@ -430,30 +481,65 @@ where
...
 
 #[cfg(test)]
 mod test {
+    use super::*;
+
+    #[cfg(debug_assertions)]
+    mod debug {
+        use super::*;
+
+        #[test]
```
pchampin:
```suggestion
        #[cfg(debug)]
        #[test]
```
The behaviour tested by this function is only valid in debug mode.
If you ran `cargo test --release`, it would fail.

pchampin:
> In the last commit I added a `resolve_triples()` method to `TripleSource`

As [previously stated](https://github.com/pchampin/sophia_rs/pull/49#issuecomment-593436346), I have mixed feelings about this.

On the one hand, I can see that this is a rather general-purpose method on triple/quad stream, that may deserves a dedicated method.

But on the other hand, if we go down that path, there are many other general-purposes functions that I can think of that would equally deserve their own specific method:

* rename bnodes into fresh ones
* replace bnodes with variables / variables with bnodes
* change absolute IRIs to relative ones (given a base IRI) (opposite of resolve)
* replace variables by their bound values (given a mapping)

We would end up with a rather cluttered API, while all of this can be achieved rather simply with `map_triples`. In the case of `resolve_triples`, this would be
```rust
   source.map_triples(|t| base.resolve(t))
```

sophia/src/term/iri.rs
MattesWhite: Yes, I think it is worth an issue, too. Given the current structure of `IriParsed` this is really challenging. I'm not happy the the current solution either as it should work with every IRI (reference). As you already noticed I only try to get the basic stuff working in this PR and leave space for optimizations later.

sophia/src/term/iri.rs
MattesWhite: Maybe, it's no longer `unsafe` but still a thread to data consistency. What would you prefer? Warning, Contract, Note, ...?

MattesWhite:
You're right `resolve_triples()` is more or less a convenient method for `map_triples()`. And especially as it is a convenience I'd like keeping it. Besides the constraints of `Iterator`s they have a pretty rich API. And you get all of it by simply implementing `next()`. Of course most `Iterator` methods can be implemented by `map()` and `fold()` as well. Triple-/QuadSource can be the same where you get a rich API of convenience methods for free.

Another point is discoverability. With many convenient methods someone can directly see the features `sophia` provides. Whereas, you have to figure out yourself how the parts fit together when there are only a few methods.

MattesWhite:
Further digging into `*Source` adapters I found that there are many duplicate items. I don't know if this is intentional or not. For the later case a comitted a version implementing `Resolver` for a `QuadSource`. This way we could remove all the duplicated adapters from `quad::stream` and even the `StreamingMode`s should be removable.

pchampin:
re. including `resolve_triples` in `TripleSource`: ok, you convinced me :-)

re. duplicate items in `triple::stream` and `quad::stream`: first I to reuse the types from `triple::stream` in `quad::stream` (only adding additional traits to them), but this required a bunch of `pub(crate)` qualifiers, which I thought were not very elegant...

sophia/src/term/iri/_join.rs
MattesWhite: This issue was resolved in 4b84a9a

sophia/src/term/iri.rs
MattesWhite: Resolved in e02c4ed

MattesWhite:
It seems to me that the issue of adapters for Triple- and QuadSources goes of topic of this PR.

I conclude this PR with an `IntoIterator` implementation for `Resolver`.

There is no coding left I want to do regarding the PR's topic. Therefore, I removed the [WIP] tag and suggest this PR for merging.

sophia/src/triple/stream/_resolve.rs
```diff
@@ -0,0 +1,140 @@
...
+        let base = &self.base;
+
+        match source.for_some_triple(&mut |t| {
+            triple = Some(base.resolve(&t).map_err(SinkError));
+        }) {
+            Ok(true) => triple,
+            Ok(false) => None,
+            Err(e) => Some(Err(SourceError(e))),
+        }
+    }
```
pchampin: This may not always work: try_for_some_triple may invoke the function more than once... So the only safe way to convert a Triple- or QuadSource to an iterator is to use an internal buffer (see triple::stream::_iterator for an example).

pchampin:
I'm happy with this PR, **except** for the `IntoIter` impl for `Resolver` (see my comment on the code). Actually, I think that the `resolve_triples`/`Resolver` thing be treated separately. In particular, I came to the conclusion that `TripleSource::resolve_triples` should be implemented as follow:

```rust
    fn resolve_triples<'a>(self, base: IriParsed<'a>) -> MapSource<Self, XXX>
    where /* ... */
    {
        self.map_triples(move |t| base.resolve(t))
    }
```
This way, we would inherit all the `impl`s attached to `MapSource`, including `IntoIter`.

The only problem is the `XXX` in the return type of the method; we need to put there the exact type of the closure, which is not possible :-/ But I can see a solution to this: make `MapSource` more generic: instead of expecting a `FnMut(StreamedTriple<S::Triple>) -> T`, it would expect a trait `SourceMapper<S>`, which would be implemented by `FnMut(StreamedTriple<S::Triple>) -> T`, but also other types...

I suggest we get rid of all the parts of this PR that relate to `resolve_triple` and the `Resolver` type. This will be added in the future PR addressing #52. Ideally, this would be done by cherry-picking the relevant commits and submitting a clean branch. But if that's too much hassle, and you prefer to just remove all this with an additional commit, I'll understand.

sophia/src/term.rs
```diff
@@ -206,50 +209,62 @@ where
...
+    /// Create a new IRI-term from a given IRI without checking its validity.
+    ///
+    /// As it is not checked if absolute or relative this property must be
+    /// entered as well.
     ///
     /// # Safety
-    /// This function requires that `iri` is a valid IRI reference,
-    /// and that `abs` correctly indicates whether it is absolute or relative.
-    pub unsafe fn new_iri_unchecked<U>(iri: U, abs: Option<bool>) -> Term<T>
+    ///
```
pchampin: I suggest renaming this section from "Safety" to "Pre-conditions"

sophia/src/term.rs
```diff
@@ -206,50 +209,62 @@ where
...
+    /// As it is not checked if absolute or relative this property must be
+    /// entered as well.
     ///
     /// # Safety
-    /// This function requires that that
-    /// `ns` and `suffix` concatenate to a valid IRI reference,
-    /// and that `abs` correctly indicates whether it is absolute or relative.
-    pub unsafe fn new_iri2_unchecked<U, V>(ns: U, suffix: V, abs: Option<bool>) -> Term<T>
+    ///
```
pchampin: Same here: I suggest renaming this section from "Safety" to "Pre-conditions"

sophia/src/term.rs
```diff
@@ -275,21 +290,11 @@ where
...
     {
-        if let Iri(dt) = dt {
-            Literal(T::from(txt), Datatype(dt))
-        } else {
-            panic!(format!(
-                "new_literal_dt_unchecked expects Term::Iri as dt, got {:?}",
-                dt
-            ))
-        }
+        Literal(T::from(txt), Datatype(dt.try_into().unwrap()))
```
pchampin: The `unwrap()` in this line may panic, so why removing the related section in the documentation?
```
    ///
    /// # Panics
    /// Panics if `dt` is not an IRI.
```

sophia/src/term/_literal_kind.rs
```diff
@@ -1,54 +1,55 @@
...
 pub use self::LiteralKind::*;
 
 impl<T> LiteralKind<T>
 where
-    T: AsRef<str>,
+    T: TermData,
 {
     /// Copy another literal kind with the given factory.
-    pub fn from_with<'a, U, F>(other: &'a LiteralKind<U>, mut factory: F) -> LiteralKind<T>
+    pub fn from_with<'a, U, F>(other: &'a LiteralKind<U>, mut factory: F) -> Self
```
pchampin: Shouldn't this be renamed as well?
```suggestion
    pub fn copy_with<'a, U, F>(other: &'a LiteralKind<U>, mut factory: F) -> Self
```

sophia/src/term.rs
```diff
@@ -206,50 +209,62 @@ where
         }
     }
 
-    /// Copy another term with the given factory,
-    /// applying the given normalization policy.
-    pub fn normalized_with<U, F>(other: &'_ Term<U>, mut factory: F, norm: Normalization) -> Term<T>
+    /// Transforms the underlying IRIs according to the given policy.
+    ///
+    /// If the policy already applies the Term is returned unchanged.
+    pub fn normalized(&self, policy: Normalization) -> Cow<'_, Self>
```
pchampin: I'm not a big fan of this change of API. Maybe my intent would have been clearer if I had named the method `copy_normalized_with`. The main goal was to *copy* the term, while ensuring a given "internal form" for IRIs...

This change means that `TermFactory::copy_normalized` has been commented out, which I think is a shame...

sophia/src/term/iri.rs
pchampin: As elsewhere, I suggest `Pre-conditions`

sophia/src/term/iri/_regex.rs
```diff
@@ -1,205 +1,27 @@
...
-            path.remove(i);
-        } else {
-            i += 1;
-        }
-    }
-}
-
 lazy_static! {
-    static ref IRI_REGEX: Regex = Regex::new(r"(?x)^
+    /// Match an IRI reference either absolute or relative.
```
pchampin:
```suggestion
    /// Match an absolute IRI reference
```
it does *not* match relative IRI references.

Note that, according to RFC3987, an IRI is always absolute. An **IRI reference**, on the other hand, can be absolute or relative (and an absolute IRI reference happens to be an IRI). That's why this regex is just called IRI_REGEX.

sophia/src/triple/stream/_resolve.rs
MattesWhite: Removed in efe2a62

sophia/src/term.rs
MattesWhite: Your right, re-included documentation (don't know why I removed it in the first place...)

sophia/src/term/iri/_regex.rs
MattesWhite: Corrected documentation and added a note to documentation of `sophia::term::iri` about the naming convention.

I did not accept this as a commit to reduce absolute number of commits.

sophia/src/term/_literal_kind.rs
MattesWhite: I plan to do so but latter in the PR `Literal as own public type` :wink:.

BTW as suggested by you (https://github.com/pchampin/sophia_rs/pull/35#issuecomment-577721198) `LiteralKind` will then be removed anyway.

sophia/src/term.rs
MattesWhite: Okay, I forgot to refactor that method in `TermFactory` :+1: .

Furthermore, I see now the use-case for `copy_normalized_with()`. I have to admit, I had a different use-case in mind. But your explanation makes sense. I'll remove this change.

MattesWhite: checkout changes in 488aaa8

sophia/src/term/iri.rs
```diff
@@ -4,6 +4,16 @@
...
 //!
+//! # Naming
+//!
+//! > According to RFC3987, an IRI is always absolute. An IRI reference, on the
+//! > other hand, can be absolute or relative (and an absolute IRI reference
+//! > happens to be an IRI).
+//!
+//! In most cases when the documentation talks about IRIs it is actually
+//! referred to IRI references. Only if IRI reference, absolute IRI or relative
+//! IRI are explicitly mentioned the difference matters.
```
pchampin:
```suggestion
//! In most cases when the documentation talks about IRIs it actually
//! refers to IRI references. Only if IRI reference, absolute IRI or relative
//! IRI are explicitly mentioned does the difference matter.
```

MattesWhite:
- [x] No further content to add to this PR.
- [x] Conversations are resolved.
- [x] Documentation is added.
- [x] Tests were added.

I think its ready to be merged.

Issue #52, opened 2020-03-09
Provide a common abstraction of TripleSource and QuadSource


MattesWhite:
This PR originates from the discussion in #49 starting with [this comment](https://github.com/pchampin/sophia_rs/pull/49#issuecomment-595701358).

# Summary

- It would be beneficial to have convenience adapters for general tasks on `*Sources`, e.g.:
  - rename bnodes into fresh ones
  - replace bnodes with variables / variables with bnodes
  - change absolute IRIs to relative ones (given a base IRI) (opposite of resolve)
  - replace variables by their bound values (given a mapping)
  - resolve IRIs given a base IRI
- The current crate structure required to duplicate much of the code for adapters (and `streaming_mode`).

# Problems

1. **Implementing traits with associated traits does not consider trait bounds.** Meaning is there is an adapter `Normalizer<S>` one can only implement either `Iterator<Item = Triple> for Normalizer<S: TripleSource>` or `Iterator<Item = Quad> for Normalizer<S: QuadSource>`. Both together would result in `conflicting implementation`.
1. **Splitted implementations violate privacy bounds.** Declaring a `struct` in `crate::triple::stream` and implementing `QuadSource` in `crate::quad::stream` requires the fields to be `pub(crate)` at least.

# Solutions

**For 1.:** Maybe some overall abstraction for statements?

**For 2.:** Merge the `stream` submodules into a top-level module `stream` equal to `triple` and `quad`.

pchampin:
> Merge the stream submodules into a top-level module stream equal to triple and quad.

I started trying this out, factorizing some of the code into a `StatementSource` trait, then defining specializations of it as `TripleSource` and `QuadSource`. But I ran into the following problem:

One can not easily use the specialized traits in a trait bound. Instead of writing
```rust
        pub fn consume_triple_source<TS: TripleSource>(mut ts: TS) 
        { /*...*/ }
````
one needs to write
```rust
        pub fn consume_triple_source<TS: TripleSource>(mut ts: TS) 
        where <TS::Mode as StreamingMode>::Item: UnsafeTriple
        { /*...*/ }
````
which is unintuitive and cumbersome. This does not go in the direction of making the API easier...

I put [a summarized version of this problem](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cb7ee5fdcb52bfff47d8f143ac7fd461) in the playground.

pchampin:
[...]

PR #53, opened 2020-03-17
Introduce Literal as an own type

MattesWhite:
Final contribution to making kinds of RDF terms own types.

Previous PRs: #49 , #48 , #47 

Closes: #34

Input from: #35 and previous PRs

---

This PR refactors `sophia::term::Term`'s `Literal` variant into an own, public type and aligns its interface to the APIs of the previous PRs for other kinds of terms.

Due to its scale and decisions to discuss this is marked work-in-progress.

**ToDo:**

- [x] Create type `Literal`.
- [x] Refactor `_literal_kind.rs` into `literal.rs`.
- [x] Integrate `Literal` into `Term`.
- [x] Refactor conversion primitive types <-> literal.
- [ ] Final alignment of `Variable`, `BlankNode`, `Iri` and `Literal` to provide a uniform API for terms.
  - [x] Remove `unsafe` from all `*_unchecked()` methods and introduce `debug_assert!()`.
  - [x] Similar implementations of `From`, `TryFrom`, `PartialEq` and other convenience traits.
  - [ ] ...

As always, the todo-list is my personal roadmap and is open to discussion.


MattesWhite:
For me I did everything on my list is done and now the PR is ready to be merged.

If any changes are requested I'm happy to follow or discuss.

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+
+use super::*;
+use crate::ns::xsd;
+use crate::term::{Iri, Result, StaticTerm, TermData, TermError};
+use crate::triple::stream::{SinkError, SourceError, StreamResult};
+
+/// Provide an IRI to specify the semantic datatype.
+pub trait DataType {
+    /// Static datatype of this Rust-type.
+    fn dt() -> &'static Iri<&'static str>;
```
pchampin: might as well be a `const`...
And I would rather call it `IRI`.

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    /// Datatype of `self`.
+    ///
+    /// This can be used for `enum`s where each variant may has its own
+    /// datatype.
+    fn dt_of(&self) -> &'static Iri<&'static str> {
```
pchampin: Might look strange when you call it with the `.` notation: `x.dt_of()`.
I suggest renaming it `dt`, or better `datatype`.

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+        Self::dt()
+    }
+
+    fn dt_term() -> StaticTerm {
```
pchampin: Rename to `iri_as_term`?

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+        Iri::from(Self::dt()).into()
+    }
+
+    fn dt_term_of(&self) -> StaticTerm {
```
pchampin: Rename to `datatype_as_term`?

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+        Iri::from(Self::dt_of(self)).into()
+    }
+}
+
+/// Representation as an RDF literal.
+pub trait AsLiteral<TD: TermData>: DataType {
```
pchampin: Does this really need to be a separate trait from `Datatype`?

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+    type Error: std::error::Error;
+
+    /// Direct from a literal.
+    fn from_literal<TD: TermData>(lit: &Literal<TD>) -> Result<Self, Self::Error>;
+
+    /// Get a term that is actually a literal.
+    ///
+    /// # Error
+    ///
+    /// The blanked implementation throws `SourceError(UnexpectedKindOfTerm)`
```
pchampin:
```suggestion
    /// The default implementation throws `SourceError(UnexpectedKindOfTerm)`
```
I think you wanted to write "blanket", but what you really meant is "default" :)

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+macro_rules! impl_dt_with_term {
+    ($ty:ty, $term:expr) => {
+        impl $crate::term::literal::DataType for $ty {
+            fn dt() -> &'static $crate::term::iri::Iri<&'static str> {
+                if let Term::Iri(iri) = &$term {
+                    iri
+                } else {
+                    panic!("Should not have passed a non IRI")
+                }
+            }
```
pchampin: I can see now why you didn't make it a `const`...
But I still think a `const` would be cleaner, and we can work around this, by declaring static `IRI`s (not `Term`s) for the relevant datatypes. This could be done by duplicating code from `ns::xsd`, or by using lazy_static of the form:
```rust
    static XSD_INTEGER = xsd::integer.try_into::<IRI>().unwrap();
```

sophia/src/term/literal/_convert.rs
```diff
@@ -0,0 +1,495 @@
...
+    };
+}
+
+impl_dt_with_term!(u8, xsd::unsignedByte);
+impl_dt_with_term!(u16, xsd::unsignedShort);
+impl_dt_with_term!(u32, xsd::unsignedInt);
+impl_dt_with_term!(u64, xsd::unsignedLong);
+impl_dt_with_term!(i8, xsd::byte);
+impl_dt_with_term!(i16, xsd::short);
+impl_dt_with_term!(i32, xsd::integer);
```
pchampin:
```suggestion
impl_dt_with_term!(i32, xsd::int);
```
`xsd:integer` is the common supertype of all types of integer. As you chose to use the most specific type for the others, it makes more sense to do the same here.

sophia/src/term/literal/_convert.rs
MattesWhite: While writing this code I had in mind having two flavors of `from_literal` traits. One that only cares about the lexical value which is currently implemented in `FromLiteral` maybe it is better called `FromLexicalValue`. The second checking the right datatype, using the `DataType` trait. However, as you mention below there is a type hierarchy in `xsd`, so I haven't figured out how to handle that best.

sophia/src/term.rs
```diff
@@ -153,31 +148,34 @@ where
         BlankNode::new(id).map(Into::into)
     }
 
+    /// Return a new simple literal with datatype `xsd:string`.
```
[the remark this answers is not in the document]
MattesWhite: ... or `AsLiteral` :wink:

sophia/src/term/literal.rs
```diff
@@ -0,0 +1,477 @@
...
+pub use self::_convert::*;
+
+/// An RDF literal.
+///
+/// Each literals has a lexical value, i.e. a text, and a datatype.
+#[derive(Clone, Copy, Debug, Eq)]
+pub enum Literal<TD: TermData> {
+    /// Simple literals are of type `xsd:string` and have shorthands in many
+    /// serialization formats.
+    Simple(TD),
```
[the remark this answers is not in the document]
MattesWhite: Okay I can understand the historical reasons why you are reluctant to this. But let me explain my intension. 

In my experience the `xsd:string` literal is quite often used. Accordingly, this (probably) requires a lot of cloning `ns::xsd::string`. Therefore, I thought it would be a nice performance gain to remove the clones. All in all, this was more a POC rather than something I insist on. In addition, having a `xsd:string` variant would mean we would require also a `xsd:integer`, `xsd:boolean`, ... and that's where things would have gotten messy.

On the other, you suggest to make `Literal` opaque which mitigates your argument of matching against `Literal::Simple`.

Well, in the end you convinced me that the `Simple` variant causes more trouble than it solves problems. I'll remove it with the next commit together with the `new()` constructor. Instead, I extend the documentation towards `AsLiteral`.

sophia/src/term/literal/_convert.rs
pchampin: Yes, the type hierarchy is a problem. Some people might prefer to convert Rust types in the most precise datatype (e.g. to ensure roundtrip), while some others might prefer to export all integer values to `xsd:integer` (for homogeneity)...

If we have to chose, I think it is better to keep the most precise information. It is stilss possible to "project" it later to more general datatypes (e.g. adding such a utility method to the list at #52).

sophia/src/term/literal.rs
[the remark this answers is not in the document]
pchampin: > I'll remove it with the next commit

Thanks :)

I can also see you point re. optimization and cloning xsd types (not just `xsd::string`) over and over again. Once `Literal` becomes opaque, we can indeed add such an optimization without risk for the users, but I would go for something more general: the datatype could be stored internally *either* as an `Iri<TD>` *or* an `Iri<&'static str>`. A constructor `new_dt_static` would allow you to pass terms from the `ns::xsd` without requiring a clone...

I'm **not** suggesting you add this before I merge the PR. But this could be an interesting improvement in the future (especially for the `AsLiteral` implementations).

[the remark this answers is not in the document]
MattesWhite: Nice to see you have the same thoughts as I do. For some time now I think about an implementation of `TermData` like:

```rust
enum MaybeStatic<TD: TermData> { // Not a good name
    Normal(TD),
    Static_(&'static str),
}
```

But as you suggest that's a matter for future PRs.

sophia/src/term/literal.rs
```diff
@@ -0,0 +1,477 @@
...
+    TD: TermData,
+{
+    /// Return a new literal with type `xsd:string`.
+    pub fn new<U>(txt: U) -> Self
+    where
+        TD: From<U>,
+    {
+        Self::Simple(txt.into())
+    }
+
```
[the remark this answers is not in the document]
MattesWhite: You're right. -> Removed in the next commit

sophia/src/term/literal.rs
```diff
@@ -0,0 +1,477 @@
...
+    pub fn txt(&self) -> &TD {
+        match self {
+            Simple(txt) | Lang { txt, .. } | Typed { txt, .. } => txt,
+        }
+    }
+
+    /// Return a copy of the IRI of the literals datatype.
+    ///
+    /// _Note:_ A language-tagged literal has always the type `rdf:langString`.
+    pub fn dt<'a>(&self) -> Iri<TD>
```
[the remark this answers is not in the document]
MattesWhite: Good idea!

I applied the requested changes but also added `try_borrow_dt(&self) -> Option<&Iri<TD>>` for use-cases like normalize or resolve where direct access to the datatype is required.

sophia/src/term/literal.rs
```diff
@@ -0,0 +1,477 @@
...
+    TD: TermData,
+{
+    fn eq(&self, other: &str) -> bool {
+        match self {
+            Simple(txt) => txt.as_ref() == other,
+            _ => false,
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
MattesWhite: Is removed with the next commit.

sophia/src/term/literal/_convert.rs
MattesWhite: The method or the trait?

pchampin: Sorry, that was ambiguous. I was suggesting to change the method `dt()` to a const `IRI`.

MattesWhite:
One moment please I'm just doing some last minor changes myself

MattesWhite:
Now I'm done :wink:

PR #56, opened 2020-03-26
Fix 'sophia_term' dependency

MattesWhite:
As 'sophia_term' is not yet published, the dependency in `sophia/Cargo.toml` throughs an error if a crate depends on `sophia`'s git. Fixed by adding a relative path.

In addition, set version of 'sophia_term' to 0.1.0 as it is a new crate.

### DIFF
--- a/sophia/Cargo.toml
+++ b/sophia/Cargo.toml
@@ -18,7 +18,7 @@ default = []
 xml = ["lazy_static", "percent-encoding", "quick-xml", "regex", "url"]
 
 [dependencies]
-sophia_term = "0.4"
+sophia_term = { version = "0.1.0", path = "../term" }
 resiter = "0.4.0"
 rio_api = { version = "0.4.1", features = ["generalized"] }
 rio_turtle = { version = "0.4.0", features = ["generalized"] }
--- a/term/Cargo.toml
+++ b/term/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sophia_term"
-version = "0.4.0"
+version = "0.1.0"
 authors = ["Pierre-Antoine Champin <pchampin@liris.cnrs.fr>"]
 edition = "2018"
 description = "A Rust toolkit for RDF and Linked Data - Types for RDF terms"


pchampin:
I see where this comes from; a student of mine also had a dependency problem when he pulled the commit splitting sophia in two distinct crates. However, I don't think this PR is the correct way to solve it.

First: I don't know how the `path` hack in `sophia/Cargo.toml` will behave once published on crates.io, but I'm guessing not very well.

Second: I wish to keep the version numbers of all sophia crates aligned, (at least on the first two components). This will make maintenance much easier (even though it might seem strange for sophia-term to begin its life at version 0.4.0... :wink:)

The clean solution to your problem, I think, is to add the following section to *your* `Cargo.toml`:
```toml
[patch.crates-io]
sophia-term = { git = 'https://github.com/pchampin/sophia_rs' }
```
This instuct cargo to look for sophia-term in the git repo, whoever asks for it (including sophia itself).

We found this solution here:
https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html

Does this solve your problem?

MattesWhite:
Ok, PRs are much faster committed then they should :wink: .

After further research we are both right. In order to use the newest version of `sophia` in `metis` I'll add the patch-section to my `Cargo.toml`. However, the `path` should also be added to `sophia/Cargo.toml`. It allows you to work with the [local version](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) of `sophia_term`during development once the (sub-)crate is published.

Regarding the version of `sophia_term` I thought this might be a copy'n'paste error. As this is intentional I will not argue about versioning.

Accordingly, you can either close this PR or I'll remove the changes at `term/Cargo.toml` in an additional commit and you can squash the commits.

pchampin:
> However, the `path` should also be added to `sophia/Cargo.toml`. It allows you to work with the [local version](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations) of `sophia_term` during development once the (sub-)crate is published.

Awesome, thanks for doing the research ;-) I'll do this right away.

pchampin:
FTR, from a few tests that I just made :
* `path = "../term"` in `sophia/Cargo.toml` solves the problem for *local* installations only (i.e. when your `Cargo.toml` specifies a `path` for sophia);
* however, if your `Cargo.toml` uses `git = "..."` (even with a *local* git repo), you seem to need the `patch.crates.io` directive.

So both solve slightly different problems.

PR #57, opened 2020-03-30
Remove resiter dependency in the tests of dataset / graphs

BruJu:
Currently we can not use the test macro for sophia graphs in other crates that implements this trait.

This first PR purpose is to remove the reciter dependency in tests.

- If the reciter dependency is kept, users have to include resiter in the Cargo.toml file which is not user friendly.
- Tests now panics when a result is an error instead of discarding the error because an error means the whole dataset was not properly retrieved (the test fails earlier)

### DIFF
--- a/sophia/src/dataset/test.rs
+++ b/sophia/src/dataset/test.rs
@@ -91,7 +91,13 @@ pub fn populate_nodes_types<D: MutableDataset>(d: &mut D) -> MDResult<D, ()> {
     Ok(())
 }
 
-pub fn as_box_q<Q: Quad>(quad: Q) -> ([BoxTerm; 3], Option<BoxTerm>) {
+pub fn as_box_q<Q: Quad, E>(quad: Result<Q, E>) -> ([BoxTerm; 3], Option<BoxTerm>) 
+{
+    let quad = match quad {
+        Ok(q) => q,
+        Err(_) => panic!("as_box_q received an error")
+    };
+    
     (
         [quad.s().into(), quad.p().into(), quad.o().into()],
         quad.g().map(|n| n.into()),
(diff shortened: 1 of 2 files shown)


sophia/src/dataset/test.rs
```diff
@@ -91,7 +91,13 @@ pub fn populate_nodes_types<D: MutableDataset>(d: &mut D) -> MDResult<D, ()> {
     Ok(())
 }
 
-pub fn as_box_q<Q: Quad>(quad: Q) -> ([BoxTerm; 3], Option<BoxTerm>) {
+pub fn as_box_q<Q: Quad, E>(quad: Result<Q, E>) -> ([BoxTerm; 3], Option<BoxTerm>) 
+{
+    let quad = match quad {
+        Ok(q) => q,
+        Err(_) => panic!("as_box_q received an error")
```
MattesWhite: As a little improvement I would suggest:
```rust
Err(e) => panic!("as_box_q received an error: {}", e)
```

sophia/src/graph/test.rs
```diff
@@ -78,7 +78,12 @@ pub fn populate_nodes_types<G: MutableGraph>(g: &mut G) -> MGResult<G, ()> {
     Ok(())
 }
 
-pub fn as_box_t<T: Triple>(triple: T) -> [BoxTerm; 3] {
+pub fn as_box_t<T: Triple, E>(triple: Result<T, E>) -> [BoxTerm; 3] {
+    let triple = match triple {
+        Ok(t) => t,
+        Err(_) => panic!("as_box_t received an error")
```
MattesWhite: Same as for `as_box_q`.

sophia/src/dataset/test.rs
pchampin: this would require that `E` implement `Display` (or `Debug` if we replaced `{}` with `{:?}`, which would probably be better).

In fact, this `panic!` was added by @BruJu to keep the function as general as possible, i.e. *not* constraining `E` to implement `Debug`. Otherwise, the whole `match` could have been replaced with a simple `unwrap()`.

Granted, error types should always implement `Debug`, so this care for genericity might be overzealous... But this is just a test function.

pchampin:
Included in PR #59, which has been merged

Issue #60, opened 2020-03-30
Handling intermediate Strings


MattesWhite:
In some situations intermediate `String`s are allocated and then transformed into `TermData`, e.g. normalization and resolving of IRIs.

# Issue

I'm currently working on `metis` for an example for #55. I'd like to use `CowTerm` for my parser. This means that if an absolute IRI is parsed I have `Cow::Borrowed` and I would like it to remain `Cow::Borrowed` after resolving against the base IRI. On the other hand when a relative IRI is parsed the `Cow::Borrowed` should be turned into `Cow::Owned` after resolution. This means that I have to know if a `&str` comes from the original to track lifetimes or if I get a newly allocated `String`.

# Discussion of Solutions

1. Change signature of `Resolve<Iri>` (and of normalization) to:
    ```rust
    impl<'a, 'td, TD, TD2> Resolve<Iri<TD>, Iri<TD2>> for IriParsed<'a>
    where
        TD: 'td + TermData,
        TD2: TermData + From<&'td str> + From<String>,
    { ... }
    ```
    This, however, prevents passing in a `TermFactory`. So maybe it would be nice to have another `Resolve`.
1. Add `resolve_with()`:
    ```rust
    trait ResolveWith<S, STD, T = S, TTD = STD> {
        fn resolve_with<'td, B, O>(&self, other: &'td S, borrowed: B, owned: O) -> T 
        where
            STD: 'td + TermData,
            B: FnMut(&'td str) -> TTD,
            O: FnMut(String) -> TTD;
    ```
    This adds complexity and is maybe to much for the single use case of `Cow` as this is not that important for other `TermData`. Besides a default implementation:
    ```rust
    /// pseudo Rust
    impl<'td, S, STD, T, TTD> Resolve<S, T> for ResolveWith<S, STD, T, TTD>
    where
        STD: 'td,
        TTD: From<&'td str> + From<String>,
    {
        fn resolve(&self, other: &S) -> T {
            self.resolve_with(other, Into::into, Into::into)
        }
    }
    ```
    Should be possible.

What do you think? Is this case important enough to add such complexity?

pchampin:
Actually, this discussion about `Term<Cow<str>>` has been on my mind lately, and I just pushed some work (39733c4a) in that spirit.

* I created my own, lighter version, of `Cow<str>`, called `MownStr` (enjoy the pun), for "maybe owned string".
* I added the `MownTerm<'a>` alias for `Term<MownTerm<'a>>`
[...]

PR #61, opened 2020-03-30
Introduce `map()` and helpers to `Term` and similars

MattesWhite:
Allows better control over ownership and transformation of 'TermData'.

The idea originated from this recent topic at URLO: https://users.rust-lang.org/t/is-it-possible-to-impl-from-wrapper-t-for-wrapper-u/40029/3

With `map()` is is possible to directly map owned `TermData`. At the moment transformation is usually done through reference via `Self<T>: From<&Self<U>>`. This rather implicit conversion can now be replaced by more explicit `self.as_ref().map(...)`. Furthermore, this can become handy for handling intermediate `String`s (#60). Another example is `term.as_ref_str()` instead of `RefTerm::from(&term)`.

---

**WIP**

At the moment this PR is more a PoC and methods are only implemented for `Iri`.

### DIFF
--- a/sophia/src/dataset/_ext_impl.rs
+++ b/sophia/src/dataset/_ext_impl.rs
@@ -56,10 +56,10 @@ impl MutableDataset for Vec<([BoxTerm; 3], Option<BoxTerm>)> {
         V: TermData,
         W: TermData,
     {
-        let s = s.into();
-        let p = p.into();
-        let o = o.into();
-        let g = g.map(|n| n.into());
+        let s = s.clone_into();
+        let p = p.clone_into();
+        let o = o.clone_into();
+        let g = g.map(|n| n.clone_into());
         self.push(([s, p, o], g));
         Ok(true)
     }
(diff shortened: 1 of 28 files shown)


pchampin:
I like the idea of `map` very much.

A few remarks
* I would allow the function passed to `map` to alter the underlying text. Doing otherwise would break the user's expectation about what `map` does (in any other type, this is the main purpose of `map` after all). Granted, modifying the string is more hazardous here, since the function does not know the *context* of the string (is it a ns? a suffix? the value of a literal?...). But I still think this sould be allowed.
*  I think we could then simply get rid of `clone_with`, which essentially does the same thing -- or maybe rename it to `map_str`, for the sake of regularity.

pchampin:
Just an afterthought: I'd like to keep `clone_with` after all, and renamingir `map_str` would *not* be a good idea: `map` (and `map_X`) consumes the term, while `clone` (and hence `clone_X`) only borrows it.

MattesWhite:
With the latest commit this PR is now ready to be merged from my perspective.

pchampin:
@MattesWhite please do not merge master in the middle of a PR. I'd rather you rebase it on top of master and force push it -- or do it myself just before the merge. It makes the history much nicer (call this vanity :wink:).

Apart from that, very nice work, thanks. I'll to the tidying and merge it.

Issue #55, opened 2020-03-24
Make 'Term' a trait


MattesWhite:
Proposed first in #35 I suggest to introduce a trait `Term`.

## Objective

`sophia`'s aim is to become a central API for RDF in Rust (#23). As such it would be nice if third party terms could smoothly interoperate with `sophia`'s ecosystem, e.g. `rio`'s terms could be stored in `sophia`'s `HashGraph`, N3 terms from `metis` could use the `Serializer` interface or another crate could provide compatible, ASCII-based terms that are based on `&[u8]` (maybe even going towards `no_std`-RDF :blush:).

## Required changes

With the latest effort to turn each kind of term (IRI, literal, blank node and variable) into an own type with an aligned API the step to create a trait is no longer that much work. A problem could be that this could turn advanced traits, such as `Graph`, in a mess of trait bounds. I'd like to commit a PoC-PR to see how it goes.

## Suggestion

Here a suggestion how a trait may look like:

```rust
trait Term: Clone + Hash + Debug + Display 
    + PartialEq + PartialEq<Iri> + PartialEq<Literal> + PartialEq<BlankNode> + Eq
where
    Iri: TryFrom<Self>,
    Literal: TryFrom<Self>,
    BlankNode: TryFrom<Self>,
{
    type TermData: TermData;

    fn value(&self) -> String;
}
```

The focus should be to keep the `Term` trait as small as possible while other behaviour is kept separate in other traits, e.g. `Resolve`.

pchampin:
I'm waiting to see your POC, but I'm not sure exactly what benefit you would gain from such a trait, especially if Iri, Literal and BlankNode are still constrained to be concrete types...

Note that nothing forces implementations of Sophia's traits (Graph, Dataset, Parser, Serializer...) to use Term internally. For example, Rio parsers do not, but still interoperate with Sophia. All that is required is to provide easy ways to convert internal "terms" to and from Sophia's Term.

For that matter, thanks to the TermData parameter, Term is already rather flexible:

* `Term<&str>` aka `RefTerm` can be cheaply constructed whenever `str` data is already available;
* self-owning variants, such as `Term<Box<tr>>` or `Term<Rc<str>>` (aka `BoxTerm` and `RcTerm` respectively) can be used in other cases;
* maybe we should also give more visibility to `Term<Cow<str>>`, giving it a type alias of its own (`CowTerm`), as it provides a nice trade-off between the two options above.

[...]

PR #62, opened 2020-04-06
Remove 'Namespace'

MattesWhite:
`Namespace` had only one functionality and was practically another `Iri`-type.

This PR removes the `Namespace` and adds `Iri::with_suffix()` which has the same semantics. 

The reason for this change is because conversion between `Namespace` and `Iri` was complicated and not necessary. In addition, improvements to `Iri` would have been required to be implemented for `Namespace` as well which no one has done. However, as `Namespace` only had a single function it was easier to move this function to `Iri`.


term/src/iri.rs
```diff
@@ -195,6 +195,38 @@ where
...
+    /// assert_eq!(&ex, "http://example.org/");
+    /// assert_eq!(&alice, "http://example.org/alice");
+    /// ```
+    pub fn with_suffix<V>(&self, suffix: V) -> Result<Self>
+    where
+        V: AsRef<str>,
+        TD: From<V>,
+    {
+        if self.suffix.is_some() {
+            Err(TermError::IsSuffixed)
```
MattesWhite, line 224: **Note on implementation:**

The `with_suffix()`-method currently throws an error if the IRI is suffixed. Theoretically, we could also use always the `Iri.ns` field and ignore `Iri.suffix`. However, I thought this would be against PolS so I decided against it.

pchampin:
Thanks for this proposal. I don't think I agree with your premise "`Namespace` was practically another Iri-type":

* a `Namespace` is neither an IRI nor a Term (you can not use it in a triple or a quad). If anything, it relates more to `TermFactory`;

* not any `Iri` can be used as a `Namespace` -- as your comment above points out, the correct behavior of `with_suffix` is not entirely trivial with "suffixed" IRIs.

Furthermore, `schema.get("Person")` looks much nicer, I think, than `Term::from(schema.with_suffix("Person"))`.

I could see the value of providing `From` conversion from `Iri` to `Namespace`, and possibly a `get_iri` method in `Namespace`, in addition to `get`... But as for replacing `Namespace` with `Iri` entirely... not convinced yet, sorry.

MattesWhite:
Okay naming `Namespace` another IRI-type were maybe not the best words. In fact, some transitions between `Namespace` and `Iri` might be more reasonable.

```rust
let var = Term::from(schema.with_suffix("Person")?);
// maybe better looking
let var: Term<_> = schema.with_suffix("Person")?.into();
```

The thing is I would really like to return `Iri` instead of `Term`. The reason is re-usability. Of course we could do a `.try_into()` if we need an IRI but this would introduce another error-handling.

---

**To give some context:**

I'm currently updating `metis` to the current state of `sophia`. This will also be the PoC of how a `Term`-trait could benefit `sophia` (see #55). I planned to use `Namespace` for the prefixes. However, after parsing a prefix's IRI it must be resolved against the current base IRI. Meaning (I) parse the IRI, (II) resolve the `Iri`  and finally (III) convert `Iri` to `Namespace`. Hence, the PR. Maybe a `impl Resolve<Namespace> for IriParsed` could also help :thinking:.

Ideally, I could reuse the prefix-parser from Turtle for N3. Which means that I would like to return an `Iri` as Turtle and N3 use different _terms_ which both implement `From<Iri>`. In fact, my N3Term looks something like:

```rust
enum N3Term<TD> {
    Iri(Iri<TD>),
    Literal(Literal<TD>),
    Existential(BlankNode<TD>),
    Universal(Variable<TD>),
    Formula(Formula<TD>), // some kind of `Graph<N3Term>`
    List(Vec<N3Term<TD>>),
}
```

As you can see I would like to reuse a `Graph` implementation of `sophia` which I can't do because it's fixed to `sophia_term::Term`.

pchampin:
> Maybe a `impl Resolve<Namespace> for IriParsed` could also help :thinking:

Yep, that would look much better to me. This and a `get_iri` method on `Namespace`, so that you can get an `Iri` instead of a `Term`...

MattesWhite:
Once again a good discussion :+1: 

I committed a new PR (#63) that applies the changes discussed. Therefore, I close this PR.

PR #65, opened 2020-04-10
simplify the Resolve trait implementation

pchampin:
This commit remove somewhat redundant implementations;
the rationale is that what they did (resolving TD1 to TD2 directly)
can usually be achieved by resolving to MownStr, then map'ing to TD2.
This will usually be just as efficient (possibly more),
because types that can recycle the memory allocated by MownStr will
usually do so (e.g. Box or String).

@MattesWhite what do you think about that. You designed and implemented the `Resolve` trait,
so I would like your opinion on this before merging it.

### DIFF
--- a/jsonld/src/test_util.rs
+++ b/jsonld/src/test_util.rs
@@ -82,6 +82,7 @@ impl<'a> Test<'a> {
             .iri()
             .parse_components(&mut buffer)
             .resolve(&test_iri)
+            .map_into()
     }
 
     pub fn positive(&self) -> bool {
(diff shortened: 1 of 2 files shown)


MattesWhite:
Nice work 👍 I think this simplification will realy increase the usability of `sophia`.

Inspecting the `_join.rs` I found the implementation of `Resolve<Literal>` a little bit akward. I think this could easily optimzed as it calls `dt.value().as_ref()`... maybe this could be improved as part of the PR. Something like matching on the literal's kind and then use `self.resolve(dt)` or so.

pchampin:
I agree that this is ugly, but I couldn't figure a way to make it work by simply resolving `other.dt()`. Lifetime hell strikes again... I don't have the bandwidth to address it right now, but fell free...

PR #66, opened 2020-04-10
Optimize 'Resolve<Literal>'

MattesWhite:
The current implementation of `impl Resolve<Literal> for IriParsed` is not optimal as it allocates intermediate string(s).

This implementation uses the standard `Resolve<Iri>` for the literal's datatype and wraps the text as `MownStr::Ref`. The key for this implementation is to move it into the `literal` module where we can access the private fields of `Literal`.

### DIFF
--- a/term/src/iri/_join.rs
+++ b/term/src/iri/_join.rs
@@ -244,36 +244,6 @@ where
     }
 }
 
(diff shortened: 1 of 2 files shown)


pchampin:
I am not a big fan of moving this particular implementation of `Resolve` away from all the others...
Turns out that I managed to solved problem inside the `_join` module (619fe8e).

PR #69, opened 2020-04-17
[WIP] Experimenting with a Term trait (as suggested in #55)

pchampin:
@MattesWhite We can use this PR to discuss about specific implementation details.

### DIFF
--- a/term/src/_trait.rs
+++ b/term/src/_trait.rs
@@ -1,14 +1,52 @@
 //! Experimenting with a `Term` trait#[derive(Eq, PartialEq)]
-
-
 use crate::mown_str::MownStr;
 
-/// A trait for types that can be used as terms
-pub trait Term: std::fmt::Debug {
+pub mod simple;
+use simple::SimpleTerm;
+
+/// A trait for types that can be used as terms in RDF graphs and datasets
+///
+/// The returned MownStr should borrow existing data as much as possible;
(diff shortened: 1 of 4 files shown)


pchampin:
I'll need some time to process you "downcasting" idea. My gut feeling is that working with `match` is more convenient, but I'll ponder about it.

I'd rather return `MownStr` than `&str` because some implementations may not have a ready to use `str` data to point to. Example: i32 as Literal. (Also, I have an idea to reduce the size of MownStr to 16 bits, just as much as `&str`, but that will be a different story).

> This way we could have implementation of the traits based on `TermData::as_str()` rather than creating intermediate `RefTerms` (or `MownTerms`).

I think my proposal also allows that. 

> Still we won't have to move the whole API of Iri and co to the traits as they are only intended to be used when terms are read-only.

I don't understand this sentence... Terms are *always* read-only, they have no mutable method...

MattesWhite:
> Terms are _always_ read-only, they have no mutable method...

Yes, I didn't flashed this out properly. What I meant was that in these cases, terms are not processed either, e.g. normalized, mapped or other methods that require access to the internals/the `TermData`. As you mentioned in #55 the `TermTrait` can/should not have constructors.

pchampin:
After further (offline) discussions, I came to agree that this proposal is too generic and not appropriate.

PR #70, opened 2020-04-21
[PoC] My term trait 

MattesWhite:
This PR implements my proposal from [my comment](https://github.com/pchampin/sophia_rs/pull/69#commitcomment-38553899) in #69.

This version adds no further types expect for the traits. The traits are implemented for the respective, existing types and an example function is given with `Graph::poc_triples_with_spo()`.

---

The contents are a _Proof of Concept_ and neither completely thought through nor completely implemented. Rather it is a base for further discussion in #69.


term/src/traits.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn as_literal(&self) -> Option<&dyn Literal> {
+        None
+    }
+    /// Downcast to a blank node.
+    fn as_blank_node(&self) -> Option<&dyn BlankNode> {
+        None
+    }
+}
+
+impl<'a> PartialEq for &'a dyn Term {
```
pchampin, line 25: Darn, I always forget we can do this (implementing a trait for a *reference* type). Nice.

MattesWhite:
Sorry for the merge misunderstood something... well this isn't going to be merged anyway...

MattesWhite:
In the latest commit I introduce `IriView` which prevents `traits::Term` from returning a trait object. But `IriView` is actually only a wrapper for `&dyn Iri` so nothing won with that solution. However, I think this could be a valuable direction to go. I try further alternatives.

MattesWhite:
As @pchampin is working on an introduction of a `Term` trait himself (which we agreed on in another online discussion), I close this PR.

PR #71, opened 2020-04-22
Change getting started example

MattesWhite:
In fact, this PR changes nothing for users but removes some boilerplate code. Therefore, it is okay to close this PR. It is rather a minor recommendation for the future.

The change uses the trick from [`rustdoc`'s documentation](https://doc.rust-lang.org/rustdoc/documentation-tests.html#using--in-doc-tests): As of Rust 1.34 we can omit the `fn main() -> Result<...>` when an `Ok::<...>(())` is returned.

_Note:_ `Box<dyn std::error::Error>` is already semantically equal to `CatchAll`.

### DIFF
--- a/sophia/src/lib.rs
+++ b/sophia/src/lib.rs
@@ -38,7 +38,6 @@
 //! use sophia::serializer::nt::NtSerializer;
 //! use sophia::triple::stream::TripleSource;
 //!
-//! # fn main() -> Result<(), CatchAll> {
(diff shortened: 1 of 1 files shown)


pchampin:
I didn't know that trick, it is great. Thanks.

And yes, `Box<dyn std::error::Error>` is better than my ad-hoc `CatchAll` hack...

Issue #72, opened 2020-04-27
Provide a generic implementation of the RDF test suite 


MattesWhite:
`sophia` provides traits for parsers and graphs. Accordingly, it should be possible to write a function that automatically runs the [RDF test suite](https://github.com/w3c/rdf-tests/) for the given format.

Following steps are required to do so:

- [x] implement a function to check if two graphs are isomorphic.
   - I think we can borrow the algorithm from [`rio`](https://github.com/Tpt/rio/blob/master/testsuite/src/isomorphism.rs) @Tpt ?
- [ ] write a framework that executes the different test cases.
- [ ] execute tests based on a `manifest`
- [ ] collect test results in a graph.
- [ ] write a function to start evaluation, e.g.
    ```rust
    fn evaluate_parser<P: Parser>(parser_factory: Fn() -> P, directory: Path) -> HashGraph
    ```
    - Of course their should also be a `evaluate_quad_parser()` function.

Except for the isomorphism everything can be feature flagged as it's normally not used.

Having the test suite available would make it easy for implementors to check their crates. In addition, we could analyse crates that have a `sophia`-wrapper and/or implement `sophia::parser::TripleParser`/`QuadParser`, e.g. `rio`.

Tpt:
I have written for Oxigraph [a better version of the isomorphism algorithm than the one provided in Rio](https://github.com/Tpt/oxigraph/blob/2bbe29c31a584ed5c2519a3d28e562b83d8a5882/lib/src/model/isomorphism.rs).

This version does not support named graphs but uses iterative hashing to try to make easier to check when there are a lot of connected blank nodes.

Feel free to reuse it, I spent more than a day on it, it would be very nice to make it useful outside of Oxigraph. Adding back named graphs support should not be hard.

MattesWhite:
I realized that the RDF test suite is rather a use-case of `sophia` than a required part of the library. Therefore, I decided that I'll publish the test suite as an own crate. Accordingly, I suppose to close this issue. 

Once the next version of `sophia` is live I'll publish `rdf-testsuite` (if I'm ready by that time :sweat_smile:). Maybe we could add a _affiliated crates_ section to `sophia`'s `README`?

pchampin:
Closing this issue, as
* it is very old
* I agree with Mattes' conclusion above: this should be a separate crate
* I have such a separate crate in the work, that I plan to publish eventually (not quite ready yet)

PR #73, opened 2020-04-29
Check Isomorphism

MattesWhite:
This PR adds a function to check whether two `Graph`s are isomorphic equivalent regarding blank nodes.

Supports: #72 

This more or less resembles the algorithm used in `oxigraph` by @Tpt. However, I had to adapt it to `sophia`'s API.

## The algorithm

To check if two RDF graphs are equal their triples must be compared. However, blank nodes are local to the scope of their graphs. Accordingly, the labels of blank nodes does not matter when we compare two graphs only that they refer to the same thing.

The idea behind the implementation is that for each BN a distinct hash is calculated. if we do this for each BN in the two graphs a mapping from BNs of the first graph to the BNs of the second graph can be derived. If the mapping is applied to the BNs of the first graph the result should be absolute equal to the second graph.

To calculate the distinct hashes for a BN we calculate the hash of the hashes of each triple that contains the BN at any position. To eliminate the dependency from BN labels, BNs are excluded from the hashing.

Should the hashes of BNs collide within a graph the hashes for those BNs are improved. This is done by hashing adjacent triples in the graph, i.e. we traverse further from subjects and objects and include the traversed triples in the hash. Iteratively the hashes are improved until no collisions are left. (In my implementation the traversal is actually limited to 8 steps). 

## ToDo

- [x] Add more (complex) tests to verify the implementation.
- [x] Implementation for datasets

## Future work

The code should become easier when we introduce a term-trait.

---

@pchampin sadly I had to add some utility function are those okay?

@Tpt could you verify that my version of the algorithm is correct?

### DIFF
--- a/sophia/src/graph/_isomorphism.rs
+++ b/sophia/src/graph/_isomorphism.rs
@@ -53,38 +53,18 @@ impl Error for AlgorithmFailure {}
 ///
 /// # Performance
 ///
-/// Gets quickly very expensive.
-///
-/// # Known issues
-///
-/// The current implementation fails if the graphs contain _redundant_ blank 
-/// nodes. Lets assume following graphs (prefix declaration omitted):
-/// 
-/// ´´´text
-/// # g1
-/// _:a :p _:b.
-/// _:c :p _:d.
-///
-/// # g2
-/// _:a2 :p _:b2.
-/// _:c2 :p _:d2.
-/// ```
-///
-/// `g1` and `g2` are clearly isomorphic equivalent. However, the current 
-/// hashing strategy would always calculate the same hashes for every triple
-/// and it is not possible to further improve the hashes. Therefore, this
-/// implementation would fail as no distinct hashes could be calculated.
+/// As this algorithm has to traverse each graph several times the algorithm
+/// gets way more expensive with bigger numbers of triples. In the same way
+/// the number of blank nodes contributes to the costs.
 pub fn isomorphic_graphs<G1, G2, E1, E2>(g1: &G1, g2: &G2) -> StreamResult<bool, E1, E2>
 where
-    E1: 'static + std::error::Error,
-    E2: 'static + std::error::Error,
+    E1: 'static + Error,
+    E2: 'static + Error,
     G1: Graph<Error = E1>,
     G2: Graph<Error = E2>,
 {
     // quick return conditions
     // -----------------------
-    dbg!("# Quick return");
-    
     // blank nodes in the respective graph
     let bns1 = g1.bnodes().source_err()?;
     let bns2 = g2.bnodes().sink_err()?;
@@ -97,7 +77,6 @@ where
     // -------------------------------------
     // - regardless of blank nodes
     // - implicitly checks that g1 and g2 have the same length
-    dbg!("# Check length and triples");
     let g1_in_g2 = check_for_equal_triples_regardless_bns(g1, g2)?;
     let g2_in_g1 = check_for_equal_triples_regardless_bns(g2, g1).map_err(StreamError::reverse)?;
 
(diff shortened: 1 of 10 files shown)


sophia/src/graph/_traits.rs
```diff
@@ -358,6 +358,9 @@ pub trait Graph {
         Ok(res)
     }
 
+    /// Number of triples stored in the graph.
+    fn len(&self) -> usize;
+
```
pchampin, line 363: The omission of a `len` method was deliberate. I didn't want to impose to all implementations that they should be able to give an exact number of triples. We can still discuss that choice, though.

The "recommended" way of estimating the number of triples would be to call `g.triples().size_hint()`. Granted, this should be documented... :)

MattesWhite: Alright, I'll fix that

MattesWhite:
@Tpt I detected an issue in the algorithm when a graph includes redundant blank nodes. The problem is closer described in the function's documentation.

sophia/src/graph/_isomorphism.rs
```diff
@@ -0,0 +1,686 @@
...
+/// As this algorithm has to traverse each graph several times the algorithm
+/// gets way more expensive with bigger numbers of triples. In the same way
+/// the number of blank nodes contributes to the costs.
+pub fn isomorphic_graphs<G1, G2, E1, E2>(g1: &G1, g2: &G2) -> StreamResult<bool, E1, E2>
+where
+    E1: 'static + Error,
+    E2: 'static + Error,
+    G1: Graph<Error = E1>,
+    G2: Graph<Error = E2>,
+{
```
pchampin, line 65: couldn't this work?
```suggestion
pub fn isomorphic_graphs<G1, G2>(g1: &G1, g2: &G2) -> StreamResult<bool, G1::Error, G2::Error>
where
    G1: Graph,
    G2: Graph,
{
```

sophia/src/triple/stream/_error.rs
```diff
@@ -87,7 +87,54 @@ where
...
 pub type StreamResult<T, E1, E2> = Result<T, StreamError<E1, E2>>;
+
+/// Extension trait for `SourceError`s.
+pub trait SourceResult<T, E1, E2>
+where
+    E1: 'static + Error,
+    E2: 'static + Error,
+{
+    /// Map an error to a `SourceError`.
+    fn source_err(self) -> StreamResult<T, E1, E2>;
```
pchampin, line 109: wouldn't `source_result` be a more appropriate name for this method?

sophia/src/triple/stream/_error.rs
```diff
@@ -87,7 +87,54 @@ where
...
+}
+
+/// Extension trait for `SinkError`s.
+pub trait SinkResult<T, E1, E2>
+where
+    E1: 'static + Error,
+    E2: 'static + Error,
+{
+    /// Map an error to a `SinkError`.
+    fn sink_err(self) -> StreamResult<T, E1, E2>;
```
pchampin, line 129: wouldn't `sink_result` be a more appropriate name for this method?

sophia/src/graph/_isomorphism.rs
```diff
@@ -0,0 +1,686 @@
...
+///
+/// # Errors
+///
+/// Both graphs can fail traversing (and this is done several times),
+/// accordingly, a `StreamError` returned where `SourceError`s originate from
+/// `g1` and `SinkError`s originate from `g2`
+///
+/// _TODO:_ In this case the notion of `source` and `sink` is not fitting.
+/// Still this requires two different error types. Maybe we should turn
+/// `StreamError` into `EitherError`?
```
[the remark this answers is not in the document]
MattesWhite: Okay, I see your point. I only added that to start the discussion. But you are right I'll remove the comment with the next commit.

sophia/src/graph/_isomorphism.rs
MattesWhite: Nice, I'll have a look at it.

sophia/src/triple/stream/_error.rs
MattesWhite: I thought about it but I decided to stay close to `map_err()` as this does nearly the same. If the `Result` yields an errors it maps/wraps it in an `SourceError`. In fact, this only replaces `.map_err(SourceError)`.

I think to write `fallible_function().source_err()?` makes it clear that if this function fails its error will be propagated as an `SourceError`. In difference, `fallible_function().map_err(SourceError)?` requires two steps of thinking:

1. Okay this will change the error-type
2. Ah so `SourceError` is a variant of `StreamError` and only wraps the given error.

pchampin: I totally agree with the added value of this trait, vs. having to explicitly call `map_err`. And the trait name is perfect for me.

Regarding the method name, I'm not entirely convinced, but I'm not entirely against it either... So I'm ok with keeping it.

sophia/src/triple/stream/_error.rs
MattesWhite: See discussion about `source_err()`

MattesWhite:
With the latest commit this PR is complete and ready to be merged.

MattesWhite:
Does anyone know/understand why Travis CI stuck?

pchampin:
> Does anyone know/understand why Travis CI stuck?

Can you elaborate?

Note that I don't have a strong feeling between Travis and Github CI. It's just that I'm no expert in either one.

MattesWhite:
Well if I look [into Travis](https://github.com/pchampin/sophia_rs/pull/73/checks?check_run_id=656064037) it states that the check started about 4 days ago and is still queued. But no particular test is being executed so far. However, I ran the tests, clippy and fmt locally so non of them should fail.

Submitting another commit just to restart Travis wrong seems to me.

pchampin:
> Well if I look [into Travis](https://github.com/pchampin/sophia_rs/pull/73/checks?check_run_id=656064037) (...)

this is actually Github's rendering of Travis's report. If you click on the link at the bottom, you are lead [there](https://travis-ci.com/github/pchampin/sophia_rs/builds/164657789) which shows that all tests have passed 4 days ago.

So it seems to be a glitch on Github's side. Could this be intentional to lead us to migrate to GH CI? :smiling_imp:

MattesWhite:
As the PR is already merged and only Travis is still doing silly things I clean up by closing this PR.

PR #75, opened 2020-06-05
RawValue as an own type

MattesWhite:
This PR introduces the type `RawValue` as discussed in #55.

### DIFF
--- a/api/src/term/_raw_value.rs
+++ b/api/src/term/_raw_value.rs
@@ -9,7 +9,7 @@ use std::hash::{Hash, Hasher};
 /// IRIs, storing both a "namespace" and a "suffix". For other kinds of term,
 /// the second part is by default `None`.
 #[derive(Debug, Copy, Clone, Eq)]
-pub struct RawValue<'a>(&'a str, Option<&'a str>);
+pub struct RawValue<'a>(pub &'a str, pub Option<&'a str>);
 
 impl<'a> RawValue<'a> {
     /// A new raw value from namespace and suffix.
--- a/api/src/term/_raw_value.rs
+++ b/api/src/term/_raw_value.rs
@@ -55,6 +55,12 @@ impl<'a> From<RawValue<'a>> for MownStr<'a> {
     }
 }
 
+impl<'a> From<&'a str> for RawValue<'a> {
+    fn from(s: &'a str) -> Self {
+        RawValue(s, None)
+    }
+}
+
 impl<'a> PartialEq for RawValue<'a> {
     fn eq(&self, other: &Self) -> bool {
         self.len() == other.len() && self.bytes().zip(other.bytes()).all(|(s, o)| s == o)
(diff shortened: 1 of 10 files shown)


MattesWhite:
From my point of view this PR is ready to be merged. If you have any requests about the changes I'll be glad to serve them.

api/src/term/_raw_value.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// _Note:_ This function uses a heuristic that if the string contains a
+    /// colon `:` this might be a schema and therefore an absolute IRI. In
+    /// fact, there is no check that `self` represents a valid IRI.
+    pub fn is_absolute(&self) -> bool {
+        match str_absolute(self.0) {
+            Absolute::Maybe => self
+                .1
+                .map(|txt| str_absolute(txt) == Absolute::Yes)
+                .unwrap_or(false),
+            abs => abs != Absolute::No,
```
pchampin, line 32: I find that `abs == Absolute::Yes` would be more readable
-- not a big fan of double-negations :wink:

Actually, I think that
```rust
    Absolute::Yes => true,
    Absolyte::No => false,
```
would be even clearer, and not that much longer...

api/src/term/_raw_value.rs
```diff
@@ -0,0 +1,158 @@
...
+    ///
+    /// To create a raw value without suffix use `From<&str>` implementation.
+    pub fn new(ns: &'a str, suffix: &'a str) -> Self {
+        RawValue(ns, Some(suffix))
+    }
+    /// If this represents an IRI, check if its an absolute one.
+    ///
+    /// _Note:_ This function uses a heuristic that if the string contains a
+    /// colon `:` this might be a schema and therefore an absolute IRI. In
+    /// fact, there is no check that `self` represents a valid IRI.
```
pchampin, line 25: The term "heuristic" implies that it could sometimes be wrong, which is not the case (under the pre-condition that the two strings *do* concatenate to a valid IRI reference). So I propose to rephrase it like that:
```suggestion
    /// _Note:_ This function merely checks that the resulting string starts
    /// with an IRI scheme. It does *not* further check the validity of the
    /// full IRI, which is supposed to be done by the TTerm implementation.
```

api/src/term/_raw_value.rs
```diff
@@ -0,0 +1,158 @@
+//! Implementation of the so-called [`RawValue`](./struct.RawValue.html).
+
+use mownstr::MownStr;
+use std::hash::{Hash, Hasher};
+
+/// A raw value is a string possibly split into two parts.
+///
+/// The second part of the raw value is intended for some implementations of
+/// IRIs, storing both a "namespace" and a "suffix". For other kinds of term,
+/// the second part is by default `None`.
```
pchampin, line 10:
```suggestion
/// the second part must be `None`.
```

api/src/term/_raw_value.rs
MattesWhite: Implemented in f94504b

pchampin:
Commit 86baca0 is actually a merge of a rebased and adapted version of this PR. Closing as "merged".

Issue #44, opened 2020-02-25
Ensure doc coverage using #![deny(missing_docs)]


MattesWhite:
`#![deny(missing_docs)]` is a macro that can be placed in any module. Using it results in compiler errors if a public element is not documented (`///`).

While this might lead to a number of 'obvious' documentations it is still a good measure to ensure a thorough documentation of the whole crate. In addition, it forces the developer to think about the users point of few on an element.

Solving this issue consists in adding `#![deny(missing_docs)]` incrementally to each and every module, until all are documented. Then, we can keep only one directive at the top-level (`lib.rs`).

pchampin:
I'm all for it, but this will take some time to please this macro:

> error: aborting due to 256 previous errors

:-/

MattesWhite:
I found out that the macro `#![deny(missing_docs)]` can not only be applied to the whole library but also to (sub-)modules. Meaning the documentation effort can be progressed iteratively.

Therefore, I propose to mark this issue 'help_wanted' and 'good_first_issue'.

yever:
I took a stab at adding doc comments: https://github.com/pchampin/sophia_rs/pull/84

pchampin:
fixed by 0c38774

PR #85, opened 2020-08-16
Fix unresolved doc links in jsonld/src/serializer.rs

yever:
The 3 links in the doc comment of
`sophia_jsonld::serializer::jsonTarget` were emitting `unresolved
link` warnings.

The broken links can be seen in: https://docs.rs/sophia_jsonld/0.6.2/sophia_jsonld/serializer/struct.JsonTarget.html

There was a typo (`JsonLdSerialiser` instead of `JsonLdSerializer`)
and the links were specified with the wrong syntax:
`[...](...)` instead of `[...]: ...`

This is related to #44 (broken links rather than missing docs).

### DIFF
--- a/jsonld/src/serializer.rs
+++ b/jsonld/src/serializer.rs
@@ -67,13 +67,13 @@ where
 
 pub type Jsonifier = JsonLdSerializer<JsonTarget>;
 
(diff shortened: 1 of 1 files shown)


pchampin:
Well spotted. Thx

Issue #19, opened 2019-12-10
Add SPARQL support


pchampin:
- [ ] add a (set of) type(s) for representing a [SPARQL abstract query](https://www.w3.org/TR/sparql11-query/#sparqlDefinition) 
- [ ] add a parser for building a SPARQL abstract query (see above) from a string in the SPARQL syntax
- [ ] incrementally implement the SPARQL algebra (this item should probably be split into sub-items)

NB: since sophia uses a *generalized* RDF model (including variables), a [Graph](https://docs.rs/sophia/0.3.0/sophia/graph/trait.Graph.html) can also be used as a [basic graph pattern](https://www.w3.org/TR/sparql11-query/#BasicGraphPatterns). The [query](https://github.com/pchampin/sophia_rs/blob/master/sophia/src/query.rs) module contains a preliminary implementation of this idea.

Tpt:
For Oxigraph I built a SPARQL parser and a SPARQL algebra representation.

Algebra: https://github.com/Tpt/oxigraph/blob/master/lib/src/sparql/algebra.rs
Parser: https://github.com/Tpt/oxigraph/blob/master/lib/src/sparql/sparql_grammar.rustpeg
Parser invocation: https://github.com/Tpt/oxigraph/blob/master/lib/src/sparql/parser.rs

It might be interesting to build it as a separated crate and make Sophia and Oxigraph depend on it, just like Rio.

The parser is a bit slow at the moment, I am planning to rewrite it using a more efficient parsing library, probably nom. But I plan to make a working 0.1 release of Oxigraph first.

pchampin:
Thanks @Tpt for chiming in.

> It might be interesting to build it as a separated crate and make Sophia and Oxigraph depend on it, just like Rio.

I have considered this. But as I mentioned above, in sophia it would be more natural to reuse the `graph::Graph` to represent basic graph patterns, so this might not be the smoother way to go... I'm still open to ideas, though. The more work we can mutualize, the better.

Tpt:
> The more work we can mutualize, the better.

Huge +1. We could maybe have the parser in a separated crate with a fairly cheap algebra representation. Then Sophia could expose an easy to use algebra tree on top of it and Oxigraph could build from it its query plans.

[...]

PR #90, opened 2020-08-25
Generic API for SPARQL support

pchampin:
This addresses #19.

This proposal is über-minimalistic compared to what I suggested initially, but I realized that *generic* SPARQL support is a huge can of worm (see the comment at the top of the `sparql.rs` file).

So decided to go for the least common denominator. The drawback is that anyone requiring (even slightly) advanced SPARQL support will need to rely on implementation specific methods. But I still think that this minimal API has value for interoperability.

Any comment welcome.

### DIFF
--- a/api/src/sparql.rs
+++ b/api/src/sparql.rs
@@ -4,7 +4,7 @@
 //!
 //! These traits are deliberately very generic.
 //! Specific implementations may have additional features, such as:
-//! 
+//!
 //! - preparing a query for multiple use;
 //! - setting default values for `BASE`, `PREFIX`, `FROM`, `FROM NAMED` directives,
 //!   before parsing query string;
@@ -26,17 +26,16 @@ pub trait SparqlDataset: Dataset {
     type SparqlError: Error + 'static;
 
     /// Parse and immediately execute `query`
-    fn execute<'p, P>(&self, query: &str) -> Result<Self::Result, Self::SparqlError>
+    fn query<'p, P>(&self, query: &str) -> Result<Self::Result, Self::SparqlError>
     where
-        P: PrefixMap<'p> + ?Sized,
-    ;
+        P: PrefixMap<'p> + ?Sized;
 }
 
 /// The result of executing a SPARQL query.
 pub trait SparqlResult {
     type Term: TTerm;
     /// The type of iterator returned if the query was a SELECT.
-    type Bindings: Iterator<Item=Vec<Option<Self::Term>>>;
+    type Bindings: Iterator<Item = Vec<Option<Self::Term>>>;
     /// The type of graph returned if the query was a CONSTRUCT or DESCRIBE.
     type Graph: Graph;
 
@@ -53,6 +52,7 @@ pub trait SparqlResult {
 }
 
 /// Different kinds of SPARQL results
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum SparqlResultKind {
     /// Result of a SELECT query
     Bindings,
--- a/api/src/sparql.rs
+++ b/api/src/sparql.rs
@@ -1,0 +1,63 @@
+//! Common traits for working with [SPARQL](https://www.w3.org/TR/sparql11-query/).
+//!
+//! # Design rationale
+//!
+//! These traits are deliberately very generic.
+//! Specific implementations may have additional features, such as:
+//! 
+//! - preparing a query for multiple use;
+//! - setting default values for `BASE`, `PREFIX`, `FROM`, `FROM NAMED` directives,
+//!   before parsing query string;
+//! - pre-binding variables before evaluating query;
+//! - etc...
+//!
+//! However, we do not want to impose these feature, or any subset thereof,
+//! to all implementation of Sophia.
+
+use crate::dataset::Dataset;
+use crate::graph::Graph;
+use crate::prefix::PrefixMap;
+use crate::term::TTerm;
+use std::error::Error;
+
+/// A dataset that can be queried with SPARQL.
+pub trait SparqlDataset: Dataset {
+    type Result: SparqlResult;
+    type SparqlError: Error + 'static;
+
+    /// Parse and immediately execute `query`
+    fn execute<'p, P>(&self, query: &str) -> Result<Self::Result, Self::SparqlError>
+    where
+        P: PrefixMap<'p> + ?Sized,
+    ;
+}
+
+/// The result of executing a SPARQL query.
+pub trait SparqlResult {
+    type Term: TTerm;
+    /// The type of iterator returned if the query was a SELECT.
+    type Bindings: Iterator<Item=Vec<Option<Self::Term>>>;
+    /// The type of graph returned if the query was a CONSTRUCT or DESCRIBE.
+    type Graph: Graph;
+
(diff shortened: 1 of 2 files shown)


api/src/sparql.rs
```diff
@@ -0,0 +1,63 @@
...
+use crate::term::TTerm;
+use std::error::Error;
+
+/// A dataset that can be queried with SPARQL.
+pub trait SparqlDataset: Dataset {
+    type Result: SparqlResult;
+    type SparqlError: Error + 'static;
+
+    /// Parse and immediately execute `query`
+    fn execute<'p, P>(&self, query: &str) -> Result<Self::Result, Self::SparqlError>
```
Tpt, line 29: I would call this method `query` or `execute_query` to leave room for an implementation of SPARQL update

api/src/sparql.rs
```diff
@@ -0,0 +1,63 @@
...
+    where
+        P: PrefixMap<'p> + ?Sized,
+    ;
+}
+
+/// The result of executing a SPARQL query.
+pub trait SparqlResult {
+    type Term: TTerm;
+    /// The type of iterator returned if the query was a SELECT.
+    type Bindings: Iterator<Item=Vec<Option<Self::Term>>>;
```
Tpt, line 39: This iterator does not keep track of the  variable name -> tuple position mapping that is useful to e.g. make sense of a "SELECT *" query results or serialize the query results.

Tpt (commented):
This proposal looks great! Thank you! I have two minor comments.

api/src/sparql.rs
pchampin: Yes of course, you are right. I'll fix this.

api/src/sparql.rs
pchampin: The order of the variables is accessible through `SparqlResult::variables`, and the binding iterators through `SparqlResult::bindings`. Attaching the variables to the iterator *itself* would require yet another trait, making the API more complex, without a clear (to me at last) added value.

Tpt: Indeed. I should have read the code more carefully. Sorry.

What about returning something like `Vec<&str>` or `&[&str]`? This way the implementation would not have to clone the data.

pchampin: > What about returning something like Vec<&str> or &[&str]? This way the implementation would not have to clone the data.

On the other hand, that would force *all* implementations to permanently own a copy of the variable names... Granted, that would not be such a big deal.

However, I chose to make `SparqlResult::bindings` "consume" (i.e. take ownership) of the `QueryResult`. That way, an implementation could be built to just "stream" the JSON returned by a SPARQL endpoint, and not have to store anything. So you need to get ownership of the variable names before calling `bindings()`. Hence the choice to make `variables()` return `String`s rather than `&str`s.

But that's a trade-off I am open to discuss if you think there are better options.

api/src/sparql.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// The kind of this result (depends on the type of query)
+    fn kind(&self) -> SparqlResultKind;
+    /// If the query was a SELECT, the list of selected variable names
+    fn variables(&self) -> Option<Vec<String>>;
+    /// If the query was a SELECT, an iterator over all the solutions
+    fn bindings(self) -> Option<Self::Bindings>;
+    /// If the query was an ASK, the boolean result
+    fn boolean(&self) -> Option<bool>;
+    /// If the query was a CONSTRUCT or DESCRIBE, the resulting graph
+    fn graph(&self) -> Option<Self::Graph>;
```
Tpt, line 51: Taking  `&self` and not `self` might requires most of implementations to copy data here because Self::Graph lifetime is not limited by `&self` lifetime.

pchampin: You are right :-/

I am tempted to replace `&self` by `self`. Or do you suggest another option?

Tpt: Using `self` would be perfect for Oxigraph.

api/src/sparql.rs
Tpt: Thank you for make `bindings` consume the `QueryResult`. It also seems to me that it is the best option.

After your reply, I agree with you about the `variables` method return type. Returning `Vec<String>` seems indeed the best way to be easy to implement. Thank you!

api/src/sparql.rs
pchampin: So we agree.

This raises another question, though:
since methods such as `bindings` and `graph` consume `self`, it does not make much sense for them to return an `Option`. If `None` is returned, the user can not "go back" and try something else...

Put it another way, in most (if not all) situations, the user will `unwrap` the returned option. Either they know in advance what kind of result they wil get:
```rust
    let res = rdf.query("ASK { :s :p :o: }");
    if res.boolean().unwrap() {
        // ...
    }
```
or they need to `match` on `res.kind()`
```rust
    let res = rdf.query(some_query_of_unknwon_kind);
    match res.kind() {
        Bindings => {
            handle_bindings(res.bindings().unwrap());
        }
        Graph => {
            handle_graph(res.graph().unwrap());
        }
        // ...
    }
```

So wouldn't it be more natural if all the kind-specific methods would return the expected value directly *or panic if the result has the wrong kind*? This is really an open question, I haven't made up my mind about it...

yever: Wouldn't it be possible to create different `SparqlResult` traits to represent results of the different query forms (say: `SparqlSelectResult`, `SparqlAskResult`, `SparqlConstructResult`)?

Perhaps they can all derive from a common `SparqlResult` trait, so that the return value of `SparqlDataset::query` is still a `SparqlResult` but in practice, you'll get a specific one that matches the associated type `SparqlDataset::Result`?

This way, you can only get, e.g., a `boolean` if your `SparqlResult` type also implements `SparqlAskResult`, and `SparqlAskResult::boolean` (or even `SparqlAskResult::result`) could return a `bool` rather than an `Option<bool>`.

I also wonder whether it might make more sense to replace the associated type `SparqlDataset::Result` with a generic type parameter to the `SparqlDataset::query` function, so that on the same representation of a dataset, you could do something like:

```rust
let sparql_result = my_dataset.query::<SparqlAskResult>(r"
    PREFIX foaf:    <http://xmlns.com/foaf/0.1/>
    ASK  { ?x foaf:name  "Alice" }
")?;
let bool_result = sparql_result.boolean(); // or maybe: sparql_result.result();
```

pchampin: @yever
Regarding your last proposal (type parameter in the `query` method), this would not be practical when the query is provided by the user or an external program, and the developer does not know in advance which kind of result to expect. So we need an enum at some level, on which programs can match. This is achieved by the `kind` method of my proposal.

Regarding your first proposal (different traits for the different kinds of results), it would indeed be cleaner. I was first reluctant, because it would require converting the concrete result type to a Sophia-specific enum -- by contrast, the current trait can be directly implemented by Oxigraph's [`QueryResult`](https://docs.rs/oxigraph/0.1.1/oxigraph/sparql/enum.QueryResults.html). But this conversion has probably a negligible cost compared to evaluating a SPARQL query, so... worth a try.

What do you think of #91?

yever: I really like #91.

pchampin:
Closed in favour of #91

PR #91, opened 2020-08-27
Alternative API for SPARQL support

pchampin:
This is an alternative to #90, following @yever's comments.

I think I prefer this one ;-)
@Tpt what do you think?

### DIFF
--- a/api/src/sparql.rs
+++ b/api/src/sparql.rs
@@ -85,7 +85,7 @@ where
     D: SparqlDataset + ?Sized,
 {
     /// Return the list of SELECTed variable names
-    fn variables(&self) -> Result<Vec<String>, D::SparqlError>;
+    fn variables(&self) -> Vec<&str>;
 }
 
 /// A dummy module to check that implementing these traits is actually possible
@@ -109,8 +109,8 @@ mod dummy {
         }
     }
     impl SparqlBindings<MyDataset> for MyBindings {
-        fn variables(&self) -> Result<Vec<String>, Infallible> {
-            Ok(vec!["s".to_string()])
+        fn variables(&self) -> Vec<&str> {
+            vec!["s"]
         }
     }
 
--- a/api/src/sparql.rs
+++ b/api/src/sparql.rs
@@ -5,14 +5,24 @@
 //! These traits are deliberately very generic.
 //! Specific implementations may have additional features, such as:
 //!
-//! - preparing a query for multiple use;
 //! - setting default values for `BASE`, `PREFIX`, `FROM`, `FROM NAMED` directives,
-//!   before parsing query string;
+//!   before parsing query string, or
 //! - pre-binding variables before evaluating query;
 //! - etc...
 //!
 //! However, we do not want to impose these feature, or any subset thereof,
 //! to all implementation of Sophia.
+//!
+//! # Extension point
+//!
+//! A possible way to extend these traits with additional functionalities
+//! (such as the ones described above)
+//! would be to define subtraits of `Query` with additional methods
+//! (*e.g.*`set_base`, `bind_variables`...).
+//!
+//! Sophia may define such traits in the future,
+//!
+//! Sophia
 
 use crate::term::TTerm;
 use crate::triple::stream::TripleSource;
@@ -32,7 +42,7 @@ pub trait SparqlDataset {
     /// The type representing pre-processed queries.
     ///
     /// See [`prepare_query`](#tymethod.prepare_query) for more defail.
-    type Query: Query<Error=Self::SparqlError>;
+    type Query: Query<Error = Self::SparqlError>;
 
     /// Parse and immediately execute `query`.
     ///
@@ -41,13 +51,16 @@ pub trait SparqlDataset {
     ///
     /// [`prepare_query`]: #method.prepared
     fn query<Q>(&self, query: Q) -> Result<SparqlResult<Self>, Self::SparqlError>
-    where Q: ToQuery<Self::Query>;
+    where
+        Q: ToQuery<Self::Query>;
 
     /// Prepare a query for multiple future executions.
     ///
-    /// This allows some implementation to mutualize parsing,
-    /// (or any other pre-processing step) of the query string.
-    /// There is however no guarantee on how much pre-processing is actually done.
+    /// This allows some implementation to separate parsing,
+    /// (or any other pre-processing step)
+    /// of the query string from the actual exectution of the query.
+    /// There is however no guarantee on how much pre-processing is actually done by this method
+    /// (see below).
     ///
     /// # Note to implementers
     ///
@@ -55,7 +68,7 @@ pub trait SparqlDataset {
     /// you can still use `String`, which implements the [`Query`] trait.
     ///
     /// [`Query`]: ./trait.Query.html
-    fn prepare_query(&self, query_string: &str) ->Result<Self::Query, Self::SparqlError> {
+    fn prepare_query(&self, query_string: &str) -> Result<Self::Query, Self::SparqlError> {
         Self::Query::parse(query_string)
     }
 }
@@ -77,7 +90,6 @@ impl Query for String {
     fn parse(query_source: &str) -> Result<Self, Self::Error> {
         Ok(query_source.into())
     }
-
 }
 
 /// A utility trait to allow [`SparqlDataset::query`]
@@ -108,8 +120,6 @@ where
         Q::parse(self)
     }
 }
-
-
 
 /// The result of executing a SPARQL query.
 pub enum SparqlResult<T>
@@ -168,58 +178,4 @@ where
 {
     /// Return the list of SELECTed variable names
     fn variables(&self) -> Vec<&str>;
-}
-
-/// A dummy module to check that implementing these traits is actually possible
-#[cfg(test)]
-mod dummy {
-    use super::*;
-    use crate::dataset::Dataset;
-    use std::convert::Infallible;
-
-    pub type MyTerm = crate::term::test::TestTerm<String>;
-    pub type MyQuad = ([MyTerm; 3], Option<MyTerm>);
-    pub type MyDataset = Vec<MyQuad>;
-
-    pub struct MyBindings(Box<dyn Iterator<Item = Result<Vec<Option<MyTerm>>, Infallible>>>);
-
-    impl IntoIterator for MyBindings {
-        type Item = Result<Vec<Option<MyTerm>>, Infallible>;
-        type IntoIter = Box<dyn Iterator<Item = Result<Vec<Option<MyTerm>>, Infallible>>>;
-        fn into_iter(self) -> Self::IntoIter {
-            self.0
-        }
-    }
-    impl SparqlBindings<MyDataset> for MyBindings {
-        fn variables(&self) -> Vec<&str> {
-            vec!["s"]
-        }
-    }
-
-    impl SparqlDataset for MyDataset {
-        type BindingsTerm = MyTerm;
-        type BindingsResult = MyBindings;
-        type TriplesResult = Box<dyn Iterator<Item = Result<[MyTerm; 3], Infallible>>>;
-        type SparqlError = Infallible;
-        type Query = String;
-
-        fn query<Q>(&self, query: Q) -> Result<SparqlResult<Self>, Self::SparqlError>
-        where
-            Q: ToQuery<String>,
-        {
-            match query.to_query()?.borrow().as_ref() {
-                "ASK" => Ok(SparqlResult::Boolean(true)),
-                "GRAPH" => Ok(SparqlResult::Triples(Box::new(
-                    self.clone().into_iter().map(|q| Ok(q.0)),
-                )
-                    as Self::TriplesResult)),
-                _ => Ok(SparqlResult::Bindings(MyBindings(Box::new(
-                    self.subjects()
-                        .unwrap()
-                        .into_iter()
-                        .map(|t| Ok(vec![Some(t)])),
-                )))),
-            }
-        }
-    }
 }
--- a/api/src/sparql.rs
+++ b/api/src/sparql.rs
@@ -14,13 +14,12 @@
 //! However, we do not want to impose these feature, or any subset thereof,
 //! to all implementation of Sophia.
 
-use crate::dataset::Dataset;
 use crate::graph::Graph;
 use crate::term::TTerm;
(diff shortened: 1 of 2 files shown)


api/src/sparql.rs
```diff
@@ -0,0 +1,104 @@
...
+where
+    T: SparqlDataset + ?Sized,
+{
+    Bindings(T::BindingsResult),
+    Boolean(bool),
+    Graph(T::GraphResult),
+}
+
+/// The result of executing a SPARQL SELECT query
+pub trait SparqlBindings<T: TTerm>: IntoIterator<Item = Vec<Option<T>>> {
```
Tpt, line 44: Returning Vec<Option<T>> and not Result<Vec<Option<T>> will create problems with Oxigraph and streaming SPARQL evaluators: they might want to raise an error during the iteration (e.g. because of a network or disk problem).

api/src/sparql.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Borrow the list of SELECTed variable names
+    fn variables(&self) -> &[&str];
```
Tpt, line 46: As you stated in the other PR, this requires the bindings to own the `[&str]` slice. This might not be the case of all implementations that might own a `Vec<String>` or a `Vec<Variable>`.

Tpt (commented):
This looks good and very similar to Oxigraph API design!

I just have two specific concerns.

api/src/sparql.rs
pchampin: you are right, I got carried away ;)

api/src/sparql.rs
pchampin: good point, I'll fix that

api/src/sparql.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/// The result of executing a SPARQL SELECT query
+pub trait SparqlBindings<D>:
+    IntoIterator<Item = Result<Vec<Option<D::BindingsTerm>>, D::SparqlError>>
+where
+    D: SparqlDataset + ?Sized,
+{
+    /// Return the list of SELECTed variable names
+    fn variables(&self) -> Result<Vec<String>, D::SparqlError>;
```
Tpt, line 49: I don't know if there is a usecase where this method could fail (it does not in Oxigraph).

api/src/sparql.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use crate::graph::Graph;
+use crate::term::TTerm;
+use std::error::Error;
+
+/// A dataset that can be queried with SPARQL.
+pub trait SparqlDataset {
+    type BindingsTerm: TTerm;
+    type BindingsResult: SparqlBindings<Self>;
+    type GraphResult: Graph;
```
Tpt, line 25: Returning a `Graph` requires to collect all the results and not return them in a streaming way. We could maybe use an iterator instead?

Tpt (commented):
I just started to implement this in Oxigraph to check if this design works well in this case. It works! I still have two improvement suggestions.

api/src/sparql.rs
pchampin: Good idea, a TripleSource would be more appropriate.

api/src/sparql.rs
pchampin: If we assume that some implementations do not own the variable names, then we must provision for the situation were such implementations would fail to retrieve the variable names when required to. Some may e.g. perform some IO in the method `variables()`, and this IO may fail...

That being said, the more I think about it, the more I find this objection of mine rather theoretical... Consider [the XML format for SPARQL endpoint responses](https://www.w3.org/TR/rdf-sparql-XMLres/). In order to know whether a response is a Bindings or a Boolean, implementations have to go past the `<head>` element, hence to parse the variable names. So any failure there will caught by the `query` method...

So maybe the assumption that any impl of `SparqlBindings` owns a copy of the variable names, is not so bad after all... WDYT?

yever: I think ownership and certainty about the availability of the data are two separate concerns. The client could have very certainly available data regardless of whether it is owned or borrowed.

I think it should be part of the contract that if you own a `SparqlResult` type, then the results are necessarily available. Otherwise, the query would have been considered a failure (so, either IO issues or parsing issues with the result would fail in the query itself). 

I also think it doesn't really make sense to expect IO in the `SparqlResult` implementation. If you're trying to provide an API for lazy loading the results or streaming, then this could perhaps be created by a different query function, maybe `try_query()` which would return a `Result` of `SparqlLazyResult`. I'm not sure how much of a real use-case this is.

This API has now 2 levels of indirection: starting from `query()` you have a `Result` which can be `Err` or `Ok`, and then `SparqlResult` which can be `Bindings`, `Boolean` or `Triples`. Adding another `Result` to the mix seems to me a bit too defensive.

I like this API much better than the previous one (#90) and find it good. I still think it might make sense to create an API (maybe an additional function?) that allows the user to specify the expect query form and not have to destructure the result.

These are 2 different use-cases: 
1. the client code could be getting a SPARQL query from some input (say, if it's a server that gets a SPARQL query as an http request payload without knowing what query form it gets)
2. the client code could hard code a specific query, knowing in advance the query form and probably also the variables used, etc., if it's a SELECT query. Even if it's not hard-coded, the client code could expect a specific query form.

For the second use case, I think it would be helpful to have an API for doing:
```rust
let bindings = dataset.make_specific_query::<SparqlSelectQuery>("SELECT ...")?;
```

or if you don't like generics, something like:
```rust
let bindings = dataset.select_query("SELECT ...")?;
let boolean = dataset.ask_query("ASK ...")?;
let triples = dataset.construct_query("CONSTRUCT ...")?;
```

yever (commented):
I really like this API. It's much cleaner.

api/src/sparql.rs
yever: Wouldn't it make sense to split the API between a fallible one (e.g., streaming) and an infallible one (in memory, blocking IO)?

api/src/sparql.rs
pchampin: @yever 
> I think ownership and certainty about the availability of the data are two separate concerns.

You are right. Talking about ownership was a shortcut for "being able to guarantee that the `&str` can leave long enough".

> I also think it doesn't really make sense to expect IO in the `SparqlResult` implementation.

I disagree. A `SparqlResult` implementation might be streaming on the response of a SPARQL endpoint. Another implementation (such as Oxigraph) might be streaming results from a disk-based dataset, that might fail at anypoint. 

That being said, and since this was the starting point of this comment, I am less and less convinced that `variables` should return a `Result`.

>  I still think it might make sense to create an API (maybe an additional function?) that allows the user to specify the expected query form and not have to destructure the result.

This is a different issue. See below (after commit 34a3db7).

pchampin:
@yever 
In [a comment above](https://github.com/pchampin/sophia_rs/pull/91#discussion_r478430754), you wanted to make it easier for developers to handle results whose kind is known in advance.

With the commit just above, it is now possible to write:
```rust
    for binding in d.query("SELECT * {...}")?.into_bindings() {
        ...
    }
    // or
    if d.query("ASK {...}")?.into_boolean() {
        ...
    }
```
Is that enough for you? Or do you still think additional methods (`ask_query`, `select_query`) should be added? Adding them would be straighforward, but personally, I tend to prefer a single method, and the pattern above.

api/src/sparql.rs
pchampin: Segregating the API that way would reduce reusability... The path I have chosen with Sophia is to provision for errors whenever *some* implementation may fail, and use [`Infallible`](https://doc.rust-lang.org/std/convert/enum.Infallible.html) (and eventually the [`never`](https://doc.rust-lang.org/std/primitive.never.html) type) as the error of infallible implementations.

yever:
> @yever
> In [a comment above](https://github.com/pchampin/sophia_rs/pull/91#discussion_r478430754), you wanted to make it easier for developers to handle results whose kind is known in advance.
> 
> With the commit just above, it is now possible to write:
> 
> ```rust
>     for binding in d.query("SELECT * {...}")?.into_bindings() {
>         ...
>     }
>     // or
>     if d.query("ASK {...}")?.into_boolean() {
>         ...
>     }
> ```
> 
> Is that enough for you? Or do you still think additional methods (`ask_query`, `select_query`) should be added? Adding them would be straighforward, but personally, I tend to prefer a single method, and the pattern above.

Sorry, I overlooked those `into_` methods. I'm a bit concerned about panicking... ;) 

What I would have liked is to have the `query` method itself express in its `Result` type an error such as `"This SparqlResult is not a Bindings"`.

Maybe a reasonable middle ground here is to make the `into_` methods return a `Result`, so that the client code can question-mark both:
```rust
for bindings in dataset.query("SELECT ...")?.into_bindings()? {
    ...
}
```
so that eventually the client function can join these `Result`s and collect all possible errors, including "wrong result type".

It's true that it's mostly the client code's fault if they used `into_bindings()` and got a different result type, but I think it's a nicer dev experience if this doesn't blow up your program, but rather handles politely the mistake.

api/src/sparql.rs
yever: Ok, that's reasonable. I accept :)

api/src/sparql.rs
yever: > > I also think it doesn't really make sense to expect IO in the `SparqlResult` implementation.
> 
> I disagree. A `SparqlResult` implementation might be streaming on the response of a SPARQL endpoint. Another implementation (such as Oxigraph) might be streaming results from a disk-based dataset, that might fail at anypoint.
> 
> That being said, and since this was the starting point of this comment, I am less and less convinced that `variables` should return a `Result`.

As discussed in the other thread (https://github.com/pchampin/sophia_rs/pull/91#discussion_r478484044), I think this is a question of whether the API is generic and accounts for both fallible and infallible implementations, or should the APIs be split for the two cases.

I think it's fine to have the former option. It just means that a user of an infallible implementation will have to `unwrap()` a bit.

api/src/sparql.rs
pchampin: That being said, you just gave me a crazy idea: https://users.rust-lang.org/t/rfc-idea-result-t-implementing-defer-t/47910

api/src/sparql.rs
Tpt: @pchampin 

+1 to make `variables()` infallible. Parsers should be able to parse the file beginning before allowing calls to this method. Oxigraph has implemented a XML query results parser without problem doing so.

> So maybe the assumption that any impl of SparqlBindings owns a copy of the variable names, is not so bad after all... WDYT?

Indeed. For example, in Oxigraph, I have to wrap the`QuerySolutionIter` type to make it implement `SparqlBindings` so I could easily add a `variables: Vec<String>` struct member. I guess this could also be easily done by the other implementation (result parsers...).

pchampin:
@yever 
> I'm still a bit concerned about panicking... ;)

I think of those methods as very similar to `unwrap`, i.e. you should only call them if you are 100% about the pre-condition. In doubt, you should use a `match` (or `if let`) instead.

> It's true that it's mostly the client code's fault if they used `into_bindings()` and got a different result type, but I think it's a nicer dev experience if this doesn't blow up your program, but rather handles politely the mistake.

`Result` adds a little overhead, which is acceptable to handle errors that may be caused by unexpected data or runtime conditions. It should not be used, IMO, to try and catch developer's mistakes -- those must be detected beforehand, and therefore should not need any handling at runtime.

yever:
I was convinced by your last comment. I created a pull request into this pull request with some tests (#92).

I wanted to suggest adding a reference to the SPARQL issue (#19) also to this pull request.

pchampin:
Sorry for keeping this stalling for so long.
I pushed another commit with
* more documentation
* a couple of new traits for prepared queries, with a different approach.

I was reluctant to abandon totally the idea that queries could be pre-processed once and reused multiple times... This proposal makes it possible, but does not impose on implementers to do anything about it (String provides a fallback implementation for Query).

@Tpt, this notion of "prepared query" is quite different from the one in Oxigraph, but I think they can cohabit. It seems to me that `oxigraph::sparql::Query` would be a good candidate for it, but if I am wrong, you can still use `String` as a fallback.

What do you guys think?

Tpt:
> @Tpt, this notion of "prepared query" is quite different from the one in Oxigraph, but I think they can cohabit. It seems to me that oxigraph::sparql::Query would be a good candidate for it.

Yes, indeed, `oxigraph::sparql::Query` is a great candidate for it. I was also planning to remove from Oxigraph 0.2 "prepared query" that does not do much outside parsing to only use this `oxigraph::sparql::Query` type. I guess we are moving in the same direction.

api/src/sparql.rs
```diff
@@ -16,19 +16,101 @@
...
+    /// (or any other pre-processing step) of the query string.
+    /// There is however no guarantee on how much pre-processing is actually done.
+    ///
+    /// # Note to implementers
+    ///
+    /// If it is impossible or inconvenient to provide a type for pre-parsed queries,
+    /// you can still use `String`, which implements the [`Query`] trait.
+    ///
+    /// [`Query`]: ./trait.Query.html
+    fn prepare_query(&self, query_string: &str) ->Result<Self::Query, Self::SparqlError> {
```
yever, line 58: Minor linting comment: I think there should be a space after `->`

api/src/sparql.rs
```diff
@@ -16,19 +16,101 @@
...
+    /// `query` is usually either a `&str` that will be parsed on the fly,
+    /// or a `Self::Query` that was earlier prepared by the [`prepare_query`] method.
+    ///
+    /// [`prepare_query`]: #method.prepared
+    fn query<Q>(&self, query: Q) -> Result<SparqlResult<Self>, Self::SparqlError>
+    where Q: ToQuery<Self::Query>;
+
+    /// Prepare a query for multiple future executions.
+    ///
+    /// This allows some implementation to mutualize parsing,
```
yever, line 48: I'm not sure I understand what you mean by "mutualize".

yever:
I like the added docs and the support for prepared queries.

api/src/sparql.rs
pchampin: you are right. I'll run cargo fmt before merging

api/src/sparql.rs
pchampin: What I meant was "parse once, execute many times", hence the idea of "mutualizing".
But I agree that this is not the best choice of word. I'll change that as well.

PR #93, opened 2020-08-29
Removes not useful print in test code

Tpt:
Avoids to get some textual output even if all tests pass.


pchampin:
Is there any specific reason for avoid passing tests to print anything?

Tpt:
No strong reason, just avoiding cluttering the tests output with random messages.
Here, I get
```
(4, Some(4))
(0, None)
```
in the Oxigraph test output. I spent some time looking for what could print that in Oxigraph. I find it very disturbing.

pchampin:
Oh, I didn't realize this was part of the test macro, and could "pollute" someone else's test suite. Sorry about that...

PR #94, opened 2020-09-02
Use '&dyn TTerm' to access entries in a 'HashSet'

MattesWhite:
This PR adds implementations to `Term` and `dyn TTerm` so that it is now possible to get values from a `HashSet<Term<TD>>` (or `HashMap`) by using `&dyn TTerm` references, e.g.:

```rust
fn main() -> Result<(), Box<dyn std::error::Error>> {
    let t: Term<String> = Term::new_iri("http://example.com/test")?;
    let mut map = HashSet::new();
    map.insert(t);

    let iri = SimpleIri::new("http://example.com/", Some("test"))?;
    assert!(map.get(iri.as_dyn()).is_some());
    Ok(())
}
```

This also works on other implementation of `TTerm` as long as they implement `Borrow<dyn TTerm + 'a>`. However, the contract of `Borrow` requires that each type that implements the `Borrow` trait has an equivalent implementation of `Hash` and `Eq` to the one of `dyn TTerm`, i.e. they must use the `term_hash()` and `term_eq()` functions. So this is not desirable for all implementors of `TTerm`.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -32,4 +32,5 @@ lazy_static = { version = "1.4.0", optional = true }
 [dev-dependencies]
 lazy_static = { version = "1.4.0" }
 sophia_iri = { version = "0.6.2", path = "../iri", features = ["test_data"] }
+sophia_term = { version = "0.6.2", path = "../term" }
 test-case = "1.0.0"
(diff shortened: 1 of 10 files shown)


pchampin (approved):
Cool improvement thanks.
Do you consider it ready for merging?

MattesWhite:
Now it is ready to be merged :+1:

PR #105, opened 2021-11-13
Makes sample code usable as-is

hoijui:
This allows to:

1. run `cargo init`
2. add sophia to Cargo.toml
3. c&p this code into `src/main.rs`

... and be good to go.

To me, it would make sense to hide some of it in further examples,
if there were any, but at least one should be very easy to use and do first experiments with.
It certainly helps rust newbs saving a ton of time,
for the penalty of adding ~ four lines of standard code.

### DIFF
--- a/sophia/src/lib.rs
+++ b/sophia/src/lib.rs
@@ -27,36 +27,42 @@
 //! back.
 //!
 //! ```
+//! extern crate sophia;
+//! 
 //! use sophia::graph::{*, inmem::FastGraph};
 //! use sophia::ns::Namespace;
 //! use sophia::parser::turtle;
 //! use sophia::serializer::*;
 //! use sophia::serializer::nt::NtSerializer;
 //! use sophia::triple::stream::TripleSource;
-//!
-//! let example = r#"
-//!     @prefix : <http://example.org/>.
-//!     @prefix foaf: <http://xmlns.com/foaf/0.1/>.
-//!
-//!     :alice foaf:name "Alice";
-//!            foaf:mbox <mailto:alice@work.example> .
-//!
-//!     :bob foaf:name "Bob".
-//! "#;
-//! let mut graph: FastGraph = turtle::parse_str(example).collect_triples()?;
-//!
-//! let ex = Namespace::new("http://example.org/")?;
-//! let foaf = Namespace::new("http://xmlns.com/foaf/0.1/")?;
-//! graph.insert(
-//!     &ex.get("bob")?,
-//!     &foaf.get("knows")?,
-//!     &ex.get("alice")?,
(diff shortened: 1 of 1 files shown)


pchampin:
Thanks for this PR. I totally agree that it should be easy for beginners to start experimenting with Sophia. That being said, I am not 100% this is the good place for that.

Some times ago @huhn511 made a PR to include a use-documentation for Sophia, generated with mdBook. This books is currently very(!) minimalistic, but it contains a [Getting started](https://pchampin.github.io/sophia_rs/01_getting_started.html) section, containing almost exactly the code you suggest. I believe this is a better place for such a self-contained example. Would you agree?

Of course, the book needs to be accessible, which was not the case (bad GH-pages configuration... this is now fixed), and should be linked to from the README in order to be findable.

hoijui:
ahh yeah indeed, same; I am fine with it being there, so once/if it is accessible as you say, then you can close this PR.
and yeah.. very minimal.. it is basically only this one example. :D

hoijui:
Maybe an easy head-start would be, to link to some crates/software that uses sophia, and has easily accessible code (without having to read many functions in many files).
... if you know of any such.

hoijui:
Comming back to sophia again after some time, and.. it is still terrible to get started. ;-)

* in the README, there is no sample code, an therefore the same for creates.io
* on docs.rs there is one block of sample code that deserializes and serializes RDF, but it does not iterate over triples
* If one manages to find this: https://docs.rs/sophia/latest/sophia/graph/trait.Graph.html#tymethod.triples
  there is minimal code for iterating, but when it comes to doing anything with the triple, one is again lost: `// do something with t`
  -> how?

hoijui:
ouh well, to document it for myself for next time:

```
for t in graph.triples() {
    let t = t?; // rethrow error if any
    // do something with t:
    println!("{}", t.s()); // prints the subject
}
```

(I found it in my own code in an other software ... what a memory!)

pchampin:
Thanks for your renewed interest for Sophia, despite the absence of introductory material :)
Please notice that the `main` branch is now tagged as v0.8.0-alpha, which introduces a number of signifcant changes in the API. This refactoring has been taking most of the time I could dedicate to Sophia in 2022, hence the stand-by on documentation.

The [book](https://pchampin.github.io/sophia_rs/) (generated from the [`book`](https://github.com/pchampin/sophia_rs/tree/main/book/src) folder) is still in a very preliminary state, but any PR to flesh it out are welcome. As soon as it begins to contain enough material, I will link it from the README and the API documentation.

pchampin:
Closing this PR. As discussed above, the book is a better place for this kind of material.

PR #147, opened 2023-11-17
feat: add devcontainer configuration

damooo:
Adds  basic devcontainer configuration. It will allow contributors to use codespaces, and vscode remote containers.


pchampin:
I am not familiar with devcontainers, but this looks like a good idea.

Issue #143, opened 2023-11-06
Take loader_factory instead of loader in `JsonldOptions`


damooo:
Current [`JsonLdOptions<L>`](https://docs.rs/sophia_jsonld/0.8.0-alpha.3/sophia_jsonld/options/struct.JsonLdOptions.html) takes a loader of type `L`, and wraps it in a `Mutex`. It effectively makes jsonld parser non concurrent, as every concurrent op have to wait anyway till others complete.

Instead, it will be great to take a loader-factory `fn() => L`, that can return fresh loader on every call. As most loaders can be cheap to clone or construct, it will enable concurrency without much cost.

pchampin:
You are absolutely right, wrapping loaders (which have an async API) into a (sync) Mutex is an aberration. My bad!

However, I'm not entirely convinced by your assertion that loaders are "usually cheap to clone or construct". Some loaders may maintain an in-memory copy of frequently used (and possibly big) contexts. Similarly, HTTP loaders should rely on some form of caching (which could be on disk or in memory). Copying those in-memory caches would be inefficient. The alternative would be to resort to some kind of mutex *inside* the loaders...

So I would keep the idea of wrapping the loader in a mutex, but use [futures-util's async `Mutex`](https://docs.rs/futures-util/latest/futures_util/lock/struct.Mutex.html) instead. Would that make sense?

Could you maybe make a PR with these changes?

damooo:
> However, I'm not entirely convinced by your assertion that loaders are "usually cheap to clone or construct". Some loaders may maintain an in-memory copy of frequently used (and possibly big) contexts. Similarly, HTTP loaders should rely on some form of caching (which could be on disk or in memory). Copying those in-memory caches would be inefficient.

The rust pattern for this is to use interior mutability, wrapped in an `Arc` which is cheap to clone. For example, The moka's [in-memory cache](https://docs.rs/moka/latest/moka/future/struct.Cache.html), is cheap to clone, and all clones are backed by same effective cache. From their docs:

> Cloning is a cheap operation for Cache as it only creates thread-safe reference-counted pointers to the internal data structures.
> Don’t wrap a Cache by a lock such as Mutex or RwLock. All methods provided by the Cache are considered thread-safe, and can be safely called by multiple async tasks at the same time. No lock is needed.

And for on disk cache, like `CaCache` too need just path to cache.

[...]

PR #151, opened 2023-12-07
refactor JsonLdOptions to take document loader factory instead of loader

damooo:
Addresses #143

### DIFF
--- a/jsonld/src/lib.rs
+++ b/jsonld/src/lib.rs
@@ -16,6 +16,7 @@ pub use options::*;
 pub mod error;
 pub use error::*;
 pub mod loader;
+pub mod loader_factory;
 pub mod parser;
 pub use parser::*;
 pub mod serializer;
--- a/jsonld/src/loader_factory.rs
+++ b/jsonld/src/loader_factory.rs
@@ -1,0 +1,111 @@
+//! I define trait for loader factories.
+//!
+
+use std::{fmt::Display, marker::PhantomData, sync::Arc};
+
+use json_ld::Loader;
+use json_syntax::Value;
+use locspan::Location;
+use sophia_iri::Iri;
+
+/// A trait for factory of document loaders.
+pub trait LoaderFactory {
+    /// Type of loaders this factory yields.
+    type Loader<'l>: Loader<
+            Iri<Arc<str>>,
+            Location<Iri<Arc<str>>>,
+            Output = Value<Location<Iri<Arc<str>>>>,
+            Error = Self::LoaderError,
+        > + Send
+        + Sync
+        + 'l
+    where
+        Self: 'l;
+
+    /// Type of loader error.
+    type LoaderError: Display + Send;
+
+    /// Yield a new loader.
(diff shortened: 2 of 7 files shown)


pchampin:
thanks a lot @damooo .

A few questions:

* is the `LoaderFactory` trait really needed? Could we not simply make `JsonLdOptions` generic over `<LF: Fn() -> L, L: Loader<...>> ` ?? I realize that it may have usability problems, but did you consider it?

* if `LoaderFactory` is a better option, then do we really need the two specific implementations (`DefaultLoaderFactory` and `ClosureLoaderFactory`) of this type? I would rather have a blanket implementation of `LoaderFactory` for all `F: Fn() -> L where L: Loader<...>`, so that any closure/function returning a loader would implement it. Note that `MyLoader::default` would also fall into that category.

pchampin:
all this makes a lot of sense, you convinced me :)

May I suggest, however, to change the signature of `with_loader_factory` to expend a closure/function rather than a full-fledged LoaderFactory? The methodwould wrap it in a `ClosureLoaderFactory` internally. This would improve the usability of the method, IMO.

(and people could still pass SomeLoader::default if they want -- it does not really matter at this point whether we use `ClosureLoaderFactory` or `DefaultLoaderFactory` in this case)

damooo:
> (and people could still pass SomeLoader::default if they want -- it does not really matter at this point whether we use ClosureLoaderFactory or DefaultLoaderFactory in this case)

They are not the only ones? For example i will be using a *named* struct called `CloningLoaderFactory`, that have to be instantiated with a template loader, which will be cloned onwards. Then we cannot supply this factory instance?

Could we have a seperate method instead?

Also, it may be useful if we also provide named struct `CloningLoaderFactory` also by default, which covers most of the cases, so that users can name it in field types if needed.

pchampin:
> For example i will be using a _named_ struct called `CloningLoaderFactory`, that have to be instantiated with a template loader, which will be cloned onwards. Then we cannot supply this factory instance?

You could still pass the following closure:
```rust
move || templateLoader.clone()
```
and get exactly the same result, couldn't you?

damooo:
Yes, but what if we want to "name" the concrete type for generic? for type of a field in a struct?

For example, a struct, that includes parser field in it like following:

```rust
struct SomeConfig {
    jsonld_parser: JsonLdParser<CloningLoaderFactory<CachingHttpLoader>>,
   a: String,
   b: u32,
}
```

We cannot name the type of `jsonld_parser` field in the struct, as closure's type cannot be named.

pchampin:
> If we use `ClosureLoaderFactory`, we cannot name the type of `jsonld_parser` field in the struct, as closure's type is anonymous, and cannot be named.

That's right, you convinced again :)

pchampin:
I have rebased your PR (77600a2) and made a few additions (436e0e8, 54a84e6, e2d3df6), so this PR is essentially merged, even though GitHub does not recognized it. Thanks again for your work.

damooo:
_/\\\_

PR #171, opened 2024-06-26
Inline `sophia` documentation for reexported modules

mkatychev:
Current https://docs.rs/sophia/latest/sophia/ does not reexport documentation for workspace crates making documentation navigation a bit cumbersome. This PR calls `#[doc(inline)]` to re-export documentation:
https://github.com/pchampin/sophia_rs/blob/a5cca57cfad8aab51a31b24468068389b1d6616b/sophia/src/lib.rs#L27-L45

### DIFF
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -55,7 +55,6 @@ pub mod graph;
 pub mod ns;
 pub mod parser;
 pub mod prefix;
-#[doc(hidden)]
 pub mod prelude;
 pub mod quad;
 pub mod serializer;
--- a/api/src/lib.rs
+++ b/api/src/lib.rs
@@ -55,6 +55,7 @@ pub mod graph;
 pub mod ns;
 pub mod parser;
 pub mod prefix;
+#[doc(hidden)]
 pub mod prelude;
 pub mod quad;
 pub mod serializer;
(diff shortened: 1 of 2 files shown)


api/src/lib.rs
```diff
@@ -55,6 +55,7 @@ pub mod graph;
 pub mod ns;
 pub mod parser;
 pub mod prefix;
+#[doc(hidden)]
```
pchampin, line 58: Why hide it? Granted, it is not properly documented for the moment, but I think it is good to document that a prelude exists and what it contains (even if crudely).

As a reference, the standard prelude *is* included in the documentation; https://doc.rust-lang.org/std/prelude/index.html

pchampin (changes requested):

api/src/lib.rs
mkatychev: it was because the individual items in the prelude came up redundantly in the search. Perhaps it's better to hide them instead? I see your point regardless

pchampin:
Any reason for *not* including `sophia_turtle`, `sophia_jsonld` and `sophia_xml`?

mkatychev:
> Any reason for *not* including `sophia_turtle`, `sophia_jsonld` and `sophia_xml`?
> 

None other than to trigger a conversation about whether everything should be included. I'll make the changes.

pchampin:
> None other than to trigger a conversation about whether everything should be included. I'll make the changes.

IIUC, the main benefit of inlining the documentation is to allow to search all re-exported crates in one place. I think users would be surprised if only a subset of these crates was showing in the search results...

PR #172, opened 2024-07-17
Add `Send + Sync` to common traits

mkatychev:
Various error libraries have datastructures such as [`anyhow::Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html#method.new) that required the errors they hold to be `Send + Sync` in addition to being `std::error::Error`.

The common traits below have had their `::Error` associated type further constraints to include `Send + Sync`:
* `sophia_api::dataset::Dataset`
* `sophia_api::graph::Graph`
* `sophia_api::source::Source`
* `sophia_api::term::TryFromTerm`
* `sophia_inmem::index::TermIndex`

### DIFF
--- a/api/src/graph.rs
+++ b/api/src/graph.rs
@@ -66,7 +66,7 @@ pub trait Graph {
     /// The result of this method is an iterator,
     /// so it can be used in a `for` loop:
     /// ```
-    /// # fn test() -> Result<(), Box<dyn crate::Error>> {
+    /// # fn test() -> Result<(), Box<dyn std::error::Error>> {
     /// # use sophia_api::graph::Graph;
     /// # use sophia_api::term::SimpleTerm;
     /// # let graph = Vec::<[SimpleTerm;3]>::new();
@@ -85,7 +85,7 @@ pub trait Graph {
     /// # use sophia_api::graph::Graph;
     /// # use sophia_api::term::SimpleTerm;
     /// # use sophia_api::source::TripleSource;
-    /// # fn test() -> Result<(), Box<dyn crate::Error>> {
+    /// # fn test() -> Result<(), Box<dyn std::error::Error>> {
     /// # let graph = Vec::<[SimpleTerm;3]>::new();
     /// #
     /// graph.triples().for_each_triple(|t| {
@@ -114,7 +114,7 @@ pub trait Graph {
     /// # use sophia_api::prelude::*;
     /// # use sophia_api::ns::{Namespace, rdf};
     /// #
-    /// # fn test<G: Graph>(graph: &G) -> Result<(), Box<dyn crate::Error>>
+    /// # fn test<G: Graph>(graph: &G) -> Result<(), Box<dyn std::error::Error>>
     /// # where
     /// #     G: Graph,
     /// # {
@@ -137,7 +137,7 @@ pub trait Graph {
     /// # use sophia_api::ns::rdfs;
     /// # use sophia_api::term::SimpleTerm;
     /// #
-    /// # fn test<G>(graph: &G) -> Result<(), Box<dyn crate::Error>>
+    /// # fn test<G>(graph: &G) -> Result<(), Box<dyn std::error::Error>>
     /// # where
     /// #     G: Graph,
     /// # {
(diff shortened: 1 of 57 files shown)


api/src/graph.rs
```diff
@@ -53,7 +53,7 @@ pub trait Graph {
     where
         Self: 'x;
     /// The error type that this graph may raise.
-    type Error: Error + 'static;
+    type Error: Error + 'static + Send + Sync;
```
mkatychev, line 56: Knowing that this _is_ a breaking change, I held off on explicitly feature flagging unless explicitly requested. This can easily be modified:

```suggestion

    #[cfg(not(feature = "threadsafe_errs"))]
    type Error: Error + 'static;
    #[cfg(feature = "threadsafe_errs")]
    type Error: Error + 'static + Send + Sync;
```

pchampin:
Thanks for this suggestion, it makes sense. However, I would like to check first whether it breaks badly some known implementations. I will check with mine. @konradhoeffner (for `hdt`), @damooo (for `manas`) and @vemonet (for `nanopub`), could you please check whether this change would be a problem for you?

KonradHoeffner:
@pchampin: Thanks for the heads-up! In HDT I have `type Error = Infallible`, which implements Send and Sync so there wouldn't be a problem.

damooo:
Works for `manas`. For graphs and datasets, I too use Error to be Infallible, as there will be no external io. In remaining cases I explicitly mentioned `Send + Sync + 'static` bounds in where clauses.

vemonet:
Thanks for letting us know! 

I have [implemented my errors](https://github.com/vemonet/nanopub-rs/blob/main/lib/src/error.rs) following the official rust by examples docs given at https://doc.rust-lang.org/rust-by-example/error/multiple_error_types/define_error_type.html

It seems to be the recommended way to implement custom errors cleanly, and I am not a big fan of adding a whole dependency just for something as basic as handling errors (especially that the standard way to handle errors is quite nice and makes it clear which errors might be raised by the lib)

I have tried to add `impl Send for TermError {}` but it makes the compiler unhappy, it want me to make it `unsafe impl Send for TermError {}`

> the trait `std::marker::Send` enforces invariants that the compiler can't check. Review the trait documentation and make sure this implementation upholds those invariants before adding the `unsafe` keyword

Do you have an idea of how we could add `Send + Sync` to the recommended way to implement errors in rust without introducing `unsafe` code?

mkatychev:
@vemonet  I was thinking of making this a feature flag in `sophia_iri`(the "root" crate):
```
#[cfg(not(feature = "threadsafe_err"))]
pub use std::error::Error;
#[cfg(feature = "threadsafe_err")]
pub use ThreadSafeError as Error;

pub trait ThreadSafeError: std::error::Error + Send + Sync + 'static {}
impl<E> ThreadSafeError for E where E: std::error::Error + Send + Sync + 'static {}
```
then replace any `std::error::Error` instances with the `sophia_iri::Error` (wherever it's actually reexported in) above, this should make the change backwards compatible by just adding the `threadsafe_err` feature flag.

mkatychev:
I'm moving this PR to draft to implement the `threadsafe_err` feature flag discussed above

mkatychev:
@pchampin  I've implemented the `threadsafe_err` flag and fixed some with the new 1.80.0 stable that triggered some clippy errors. This PR is ready for review again

pchampin:
@vemonet
> Do you have an idea of how we could add Send + Sync to the recommended way to implement errors in rust without introducing unsafe code?

`Send` and `Sync` are special traits, which  you don't need to implement explicitly. In fact, your `TermError` and `NpError` implement them, because they do not contain any field that is `!Send` or `!Sync` (e.g. `Rc<T>`). So my take is that this change would not break your code either.

@mkatychev 
thanks a lot for the effort you put the feature flag, but in fact, I think I would rather include the change directly, without the feature flag... I'm sorry  you wasted your time :sweat_smile:. Although it is technically a breaking change, it will probably break no-one's code in practice, and the feature flag would induce a risk of splitting the ecosystem. Furthermore, we are still in 0.x versions, so breaking changes are acceptable.

Without the feature flag, I don't think that the alias `ThreadSafeError` is needed. I'd rather keep things explicit, using standard traits.

mkatychev:
@pchampin  I've reverted the flag changes, should be good to go

PR #173, opened 2024-07-24
handle `xsd::decimal` in `TryFromTerm for f64`

mkatychev:
Doing a `TryFromTerm` on a list of `xsd::decimals` would previously produce a `wrong datatype` error since the list below example would produce literals with the `decimal` type (`^^"http://www.w3.org/2001/XMLSchema#decimal"`):
```ttl
:a _:list (2.00 4.00)
```


pchampin:
Good point. Thanks for the PR

PR #174, opened 2024-07-26
Introduce `Not` for `GraphNameMatcher` and `TermMatcher`

mkatychev:
It was often found handy to exclude certain terms or graph names during a filter:
```rust
let my_graph_name: SimpleTerm<'static> =
    default_namespace.get_unchecked("#my_graph").into_term();
let ds_without_my_graph: Vec<Spog<SimpleTerm>> = dataset
    .quads_matching(Any, Any, Any, Not(Some(my_graph_name)))
    .collect_quads()
    .unwrap();
```

### DIFF
--- a/api/src/term/matcher/_any.rs
+++ b/api/src/term/matcher/_any.rs
@@ -11,3 +11,14 @@ impl TermMatcher for Any {
         true
     }
 }
+
+/// Matches on the inverse of the inner [`Term`] or [`GraphName`]
+pub struct Not<M>(pub M);
+
+impl<M: TermMatcher> TermMatcher for Not<M> {
+    type Term = SimpleTerm<'static>; // not actually used
+
+    fn matches<T2: Term + ?Sized>(&self, term: &T2) -> bool {
+        !self.0.matches(term)
+    }
+}
--- a/api/src/term/matcher/_any.rs
+++ b/api/src/term/matcher/_any.rs
@@ -11,14 +11,3 @@ impl TermMatcher for Any {
         true
     }
 }
-
-/// Matches on the inverse of the inner [`Term`] or [`GraphName`]
-pub struct Not<M>(pub M);
-
-impl<M: TermMatcher> TermMatcher for Not<M> {
-    type Term = SimpleTerm<'static>; // not actually used
-
-    fn matches<T2: Term + ?Sized>(&self, term: &T2) -> bool {
-        !self.0.matches(term)
-    }
-}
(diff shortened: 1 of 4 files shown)


api/src/term/matcher/_any.rs
mkatychev, line 1: If unit tests are needed, I'd like to know what module they should go under.

api/src/term/matcher/_any.rs
```diff
@@ -11,3 +11,14 @@ impl TermMatcher for Any {
...
+/// Matches on the inverse of the inner [`Term`] or [`GraphName`]
+pub struct Not<M>(pub M);
+
+impl<M: TermMatcher> TermMatcher for Not<M> {
+    type Term = SimpleTerm<'static>; // not actually used
+
+    fn matches<T2: Term + ?Sized>(&self, term: &T2) -> bool {
+        !self.0.matches(term)
+    }
+}
```
pchampin, line 24: it would make more sente to move this to a new `_not.rs` submodule

api/src/term/matcher/_graph_name_matcher.rs
```diff
@@ -135,3 +135,19 @@ impl GraphNameMatcher for Any {
...
+    }
+}
+
+impl GraphNameMatcher for Option<SimpleTerm<'static>> {
+    type Term = SimpleTerm<'static>;
+
+    fn matches<T2: Term + ?Sized>(&self, graph_name: GraphName<&T2>) -> bool {
+        graph_name_eq(self.as_ref(), graph_name.map(Term::as_simple))
+    }
+}
```
pchampin, line 153: I'm not comfortable with this, as Option<T> as (Graph)TermMatcher currently are used to mean "match the value of the option if any, otherwise match *nothing*". This impl would create confusion.

Note that, for any term `t` (resp. graphname `g`), wrapping it in sqare-brackets makes a matcher, so you by writing `Not([Some(my_graph_name)])` you would achieve exactly what you want, without the need to add this new impl.

pchampin (changes requested):

api/src/term/matcher/_any.rs
pchampin: all tests for (GraphName|Term)Matcher are currently in `api/src/term/matcher.rs`. Having tests for `Not` would indeed be nice.

pchampin (approved):

pchampin:
that's great; thanks

PR #180, opened 2024-10-08
chore: apply non-default clippy fixes

mkatychev:
Ran `cargo clippy --fix --all-targets --all-features -- -W clippy::uninlined_format_args -W clippy::pedantic -W clippy::nursery` to apply some non default clippy fixes automatically.

* Added `.clippy.toml` to the workspace
* Added clippy lint sectio to the workspace `Cargo.toml`

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -1,0 +1,1 @@
+allowed-wildcard-imports = [ "super" ]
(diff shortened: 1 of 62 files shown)


pchampin:
any reason why this was not applied to crates inmem and api?

mkatychev:
I don't recall, I think I had some issues with those crates triggering some false positives. I can take another look soon.

pchampin:
Thanks. I didn't know about those clippy options, but that's really an improvement :)

PR #187, opened 2025-04-01
Fix typo: "RFC 3687" -> "RFC 3987"

centromere:
The correct RFC number is [3987](https://datatracker.ietf.org/doc/html/rfc3987#section-2.2), **Internationalized Resource Identifiers (IRIs)**, not [3687](https://datatracker.ietf.org/doc/html/rfc3687), **Lightweight Directory Access Protocol (LDAP) and X.500 Component Matching Rules**.


pchampin:
Indeed, thanks!

